PR #2, opened 2023-11-19
Gql helper

JoviDeCroock:
This is meant to surface one of the potential approaches to incorporating this into front-ends.

### DIFF
--- a/src/gql.ts
+++ b/src/gql.ts
@@ -1,0 +1,60 @@
+import { IntrospectionQuery, Introspection } from './introspection';
+import { Document } from './parser';
+import { TypedDocument } from './typed-document';
+import { Variables } from './typed-document/variables';
+import { schema } from './__tests__/introspection.test-d';
+import { DocumentNode, parse } from '@0no-co/graphql.web';
+
+interface DocumentTypeDecoration<TResult, TVariables> {
+  /**
+   * This type is used to ensure that the variables you pass in to the query are assignable to Variables
+   * and that the Result is assignable to whatever you pass your result to. The method is never actually
+   * implemented, but the type is valid because we list it as optional
+   */
+  __apiType?: (variables: TVariables) => TResult;
+}
+
+interface TypedDocumentNode<Result, Variables>
+  extends DocumentNode,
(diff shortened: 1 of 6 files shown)


src/gql.ts
```diff
@@ -0,0 +1,62 @@
...
+  >(
+    document: Doc
+  ): TypedDocumentNode<Result, Vars> => {
+    return parse(document);
+  };
+};
+
+const gql = createGql<typeof schema>();
+
+// TODO: fragment type
```
JoviDeCroock, line 37: We'd need to find a way to derive something like `FragmentType` to type components easily

src/gql.ts
```diff
@@ -0,0 +1,62 @@
...
+const fragment = gql(`
+  fragment X on Todo {
+    id
+    text
+    completed
+  }
+`);
+
+// TODO: find way to flatten the fragment in here
```
JoviDeCroock, line 46: In the `Parser` we'd most likely want to include the fragments that are in the already parsed `TypedDocumentNode` - or we can always store the raw document and expand it that way

src/typed-document/index.ts
```diff
@@ -150,7 +149,7 @@ type DefinitionContinue<
   Definitions extends any[],
   Introspection extends IntrospectionType<any>,
   Fragments extends Record<string, unknown>
-> = (Definitions[0] extends OperationDefinitionNode
+> = (Definitions[0] extends { kind: 'OperationDefinition' }
```
JoviDeCroock, line 152: This fixed mutations 😅

src/gql.ts
```diff
@@ -0,0 +1,62 @@
...
+   * implemented, but the type is valid because we list it as optional
+   */
+  __apiType?: (variables: TVariables) => TResult;
+}
+
+interface TypedDocumentNode<Result, Variables>
+  extends DocumentNode,
+    DocumentTypeDecoration<Result, Variables> {}
+
+export const createGql = <Q extends IntrospectionQuery, Intro = Introspection<Q>>() => {
```
JoviDeCroock, line 21: I reckon we could also use this generic to add in custom scalars

PR #15, opened 2024-01-17
fix: Fix readFragment() not deriving complex return types

kitten:
## Summary

When a complex fragment type (e.g. any fragment that derives to a union type of multiple object shapes) is passed to `readFragment()` then the return type would only infer as the first type of the union.

```ts
// derives to `{ wallOfText: string } | { maxLength: number }`
const fragment = graphql(`
  fragment Fields on ITodo {
    ... on BigTodo {
      wallOfText
    }
    ... on SmallTodo {
      maxLength
    }
  }
`);

// previously derived to: `{ wallOfText: string }`
// rather than: `{ wallOfText: string } | { maxLength: number }`
const fields = readFragment(fragment, data);
```

## Set of changes

- Add tests for `readFragment` inputs
- Remove `Data` generic on `readFragment` and instead infer `Data` from `Document` directly

### DIFF
--- a/.changeset/neat-numbers-shake.md
+++ b/.changeset/neat-numbers-shake.md
@@ -1,0 +1,5 @@
+---
+'gql.tada': patch
+---
+
+Fix `readFragment()` not inferring the types of complex fragments, i.e. fragments that derive with a union type.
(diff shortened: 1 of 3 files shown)


JoviDeCroock (approved):
What a find

Issue #23, opened 2024-01-18
RFC: Allow to use the automatic types without fragment refs/masks


chamatt:
## Summary

Currently fragment masking is enabled by default and there's no way to disable it.

## Proposed Solution

Fragment masking introduces a lot of overhead for people that are not used to dealing with it, and could be a blocker for introducing the library into the codebase.

## Requirements

Allow to disable fragment masking, that way, a query is fully typed and fragments are unwrapped.

JoviDeCroock:
From my point of view, we are trying to encourage good patterns so I'm not sure we want to disable the fragment-masking as this pushes folks into co-locating their data-requirements with the blueprints to render said data. In establishing these patterns GraphQL clients can also more reliably make optimisations in terms of rendering bailouts or hooking into technologies like `Suspense` to enable partial UI with `@defer`.

chamatt:
Thanks for the response. Also find the patterns around fragment masking pretty interesting, sadly other folks in the team might not be too willing to make this drastic change in the way they approach graphql, especially since we're using apollo, which doesn't incentivise using this pattern.
I'll try this out some more, as it's the only typed graphGL library that worked well with our codebase (graphqlcodegen client-preset is simply too slow, TS server slows a crawl :/)

Great work you've done here!

Hebilicious:
An interesting solution that [Houdini](https://houdinigraphql.com/api/fragment#fragment-masking) uses is to introduce some special client only directives, such as @mask_disable

```gql
query CurrentUser {
	me {
		uuid
		...UserProfile @mask_disable
		...UserMeta
	}
}
```
This would disable the masking for this field only.

chamatt:
> An interesting solution that [Houdini](https://houdinigraphql.com/api/fragment#fragment-masking) uses is to introduce some special client only directives, such as @mask_disable
> 
> ```graphql
> query CurrentUser {
> 	me {
> 		uuid
> 		...UserProfile @mask_disable
> 		...UserMeta
> 	}
> }
> ```
> 
> This would disable the masking for this field only.

Love this! would be great as an opt-in approach!

kitten:
@Hebilicious: Awesome to see! I had a similar idea to disable fragment masking for `gql.tada`, so that does validate my hypothesis that it should be supported with a directive 💯

In essense, it's not that hard for us to implement.
[...]

PR #28, opened 2024-01-19
feat: allow for unmasking a fragment by means of a directive

JoviDeCroock:
Resolves https://github.com/0no-co/gql.tada/issues/23

## Summary

This adds the possibility of annotating a fragment-spread with `@mask_disable` I went for the same name as houdini for familiarity. This still needs some code to remove these directives during `parse` but wanted to already surface this PR, would we be opposed to using `visit` from `@graphql.web` for this?

### DIFF
--- a/src/selection.ts
+++ b/src/selection.ts
@@ -69,6 +69,15 @@ type isOptionalRec<Directives extends readonly unknown[] | undefined> =
       : isOptionalRec<Rest>
     : false;
 
+type isUnmaskedFragmentRec<Directives extends readonly unknown[] | undefined> =
+  Directives extends readonly [infer Directive, ...infer Rest]
+    ? Directive extends { kind: Kind.DIRECTIVE; name: any }
+      ? Directive['name']['value'] extends 'mask_disable'
+        ? true
+        : isUnmaskedFragmentRec<Rest>
+      : isUnmaskedFragmentRec<Rest>
+    : false;
+
 type getFieldAlias<Node extends FieldNode> = Node['alias'] extends undefined
   ? Node['name']['value']
   : Node['alias'] extends NameNode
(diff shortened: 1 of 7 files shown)


src/selection.ts
```diff
@@ -82,10 +91,17 @@ type getFragmentSelection<
   Fragments extends { [name: string]: any },
 > = Node extends { kind: Kind.INLINE_FRAGMENT; selectionSet: any }
   ? getSelection<Node['selectionSet']['selections'], Type, Introspection, Fragments>
-  : Node extends { kind: Kind.FRAGMENT_SPREAD; name: any }
+  : Node extends { kind: Kind.FRAGMENT_SPREAD; name: any; directives: any[] }
     ? Node['name']['value'] extends keyof Fragments
       ? Fragments[Node['name']['value']] extends FragmentDefDecorationLike
-        ? makeFragmentRef<Fragments[Node['name']['value']]>
+        ? isUnmaskedFragmentRec<Node['directives']> extends true
+          ? getSelection<
```
kitten, line 98: I think, we can probably replace `$tada.fragmentDef`, and instead of it being set to the definition, set it to either the fragment result type or `makeFragmentRef`.

This way, this part of the code doesn't have to have any knowledge of what it's embedding.

Furthermore, since `$tada.fragmentDef` would then be equivalent to `ResultOf<typeof fragment>`, this part of the code doesn't have to do _any_ work 🎉 

This does require us to place `@mask_disable` on the fragment definition, but we could do:

```graphql
fragment FragName on Type @_noMasking {
  # ...
}
```

This _then_ in turn has the benefit of `parse` only having to check the first directive and filtering the `directives` array on the `FragmentDefinition` nodes it pushes

JoviDeCroock: I thought about this and locally defined fragments already work out just fine so would this only be for fragments that we define externally outside of a component? If so this might be a tad limiting 😅 I don't mind tbh just thinking out loud

kitten: Small note, since I was mistaken,
updating `$tada.fragmentDef`’s type wouldn't be enough, since we only use the definition node via `getFragmentsOfDocumentsRec`.

So, we'd only have to update `getFragmentsOfDocumentsRec` and annotate that. It's a bit messy, since `fragmentId` is “just on there”. But we could remove `$tada.fragmentId` and replace `getFragmentsOfDocumentsRec`’s output to add a new annotation, ad-hoc.

kitten: @JoviDeCroock
> work out just fine so would this only be for fragments that we define externally outside of a component

yea, see note above, since my implementation path was flawed. We could update `getFragmentsOfDocumentsRec` to actually add the annotation that contains the type (either fragment ref or result type)

I know that this gives a lot more freedom on the spread, but I'd argue it'd have to be on the fragment definition anyway, otherwise `FragmentOf` couldn't switch to giving you the `ResultOf`-type, i.e. if we put it on the spread then the consuming component would have to have knowledge of how to define its input type.

This is of course possible, but, if we think of it as if it manipulated the runtime data, by making this configurable on the spread, there's an inconsistency between what's accepted and what's passed.

JoviDeCroock: I think I made it work 😅 it works in the example

kitten: re. docs, do you think we should write a general “Fragment Composition” page under a “Guides” section and mention it there?

I was thinking that a general “Fragment Composition” next to a future “Typed Documents” page (which would explain `TypedDocumentNode`) could explain fragment composition, masking, and component colocation.

We could also call it “Fragment Colocation” to be fair

JoviDeCroock: Hmm, should I refrain from documentation until we get the hollistic topic then?

PR #31, opened 2024-01-22
feat: Implement `@_unmask` directives to unmask fragment types

### LINKED ISSUE #23: RFC: Allow to use the automatic types without fragment refs/masks

kitten:
Supersedes #28
Resolves #23

## Summary

Fragments will now not be masked when they’re annotated with `@_unmask`. This will cause all their types derived by `FragmentOf<>` and in parent documents to be fully unwrapped and unmasked.

```ts
const fragment = graphql(`
  fragment TodoFields on Todo @_unmask {
    id
    text
  }
`);

//: { id: string | number, text: string }
type fragmentType = FragmentOf<typeof fragment>;

const query = graphql(`
  query {
    todo {
      ...TodoFields
    }
  }
`, [fragment]);

//: { todo: { id: string | number, text: string } }
type resultType = ResultOf<typeof fragment>;
```

This PR also improves the readability of the `TodoDocumentNode` generics. Instead of the third generic being a full fragment definition, it'll now be displayed with a different syntax, i.e.

```ts
{
  fragment: 'FragmentName',
  on: 'Type',
  masked: true
}
```

## Set of changes

- Refactor how reference types for definitions are used in selection types
- Update type of fragment definition (i.e. `$tada.definition`)
- Add support for `@_unmask` directive
- Filter `@_unmask` directives from fragment definitions

### DIFF
--- a/.changeset/little-dragons-repeat.md
+++ b/.changeset/little-dragons-repeat.md
@@ -2,4 +2,4 @@
 'gql.tada': minor
 ---
 
-Add support for `@_unmask` directive on fragments causing the fragment type to not be masked. `FragmentOf<>` will return the full result type of fragments when they’re annotated with `@_noMask` and spreading these unmasked fragments into parent documents will use their full type.
+Add support for `@_unmask` directive on fragments causing the fragment type to not be masked. `FragmentOf<>` will return the full result type of fragments when they’re annotated with `@_unmask` and spreading these unmasked fragments into parent documents will use their full type.


.changeset/little-dragons-repeat.md
```diff
@@ -0,0 +1,5 @@
+---
+'gql.tada': minor
+---
+
+Add support for `@_unmask` directive on fragments causing the fragment type to not be masked. `FragmentOf<>` will return the full result type of fragments when they’re annotated with `@_noMask` and spreading these unmasked fragments into parent documents will use their full type.
```
JoviDeCroock, line 5:
```suggestion
Add support for `@_unmask` directive on fragments causing the fragment type to not be masked. `FragmentOf<>` will return the full result type of fragments when they’re annotated with `@_unmask` and spreading these unmasked fragments into parent documents will use their full type.
```

JoviDeCroock (approved):

PR #32, opened 2024-01-22
feat: Implement support for `@_optional` and `@_required` on fields

### LINKED ISSUE #30: RFC: allow reacting to client-side directives
## Summary

In Relay and urql we have directives like `@_optional`/`@required` which allow us to specify how a normalized cache reacts to the absence of a given field. Something that is missing when using this in conjunction with `gql.tada` is how our types are altered in reaction to the presence of a given directive on a `field`.

## Proposed Solution

We allow a `directives` config on our `Setup` generic that instructs how to react to a given directive when found on the document, we won't strip these directives otherwise the GraphQL client won't know they were ever there but we will adjust the types.

I am not entirely sure yet how we could make this configurable, would we just need to support all of these directives in code, I assume we can't just allow for a configuration that wraps stuff in generics 😅

kitten:
Resolves #30

## Summary

This supports client-controlled nullability directives, as implemented for instance in [in Graphcache for an example.](https://formidable.com/open-source/urql/docs/graphcache/local-directives/#client-controlled-nullability)

When applied,
- `@_optional` / `@optional` turns a non-nullable field into a nullable one.
- `@_required` / `@required` turns a nullable field into a non-nullable one.

## Set of changes

- Add `getTypeDirective` as `unwrapType` input and override type as required

### DIFF
--- a/src/selection.ts
+++ b/src/selection.ts
@@ -56,9 +56,25 @@ type unwrapType<
   SelectionSet extends { kind: Kind.SELECTION_SET } | undefined,
   Introspection extends IntrospectionLikeType,
   Fragments extends { [name: string]: any },
+  TypeDirective = void,
 > = Type extends IntrospectionNonNullTypeRef
-  ? _unwrapTypeRec<Type['ofType'], SelectionSet, Introspection, Fragments>
-  : null | _unwrapTypeRec<Type, SelectionSet, Introspection, Fragments>;
+  ? TypeDirective extends 'optional'
+    ? null | _unwrapTypeRec<Type['ofType'], SelectionSet, Introspection, Fragments>
+    : _unwrapTypeRec<Type['ofType'], SelectionSet, Introspection, Fragments>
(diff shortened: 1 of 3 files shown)


src/selection.ts
```diff
@@ -56,9 +56,25 @@ type unwrapType<
   SelectionSet extends { kind: Kind.SELECTION_SET } | undefined,
   Introspection extends IntrospectionLikeType,
   Fragments extends { [name: string]: any },
+  TypeDirective = void,
 > = Type extends IntrospectionNonNullTypeRef
-  ? _unwrapTypeRec<Type['ofType'], SelectionSet, Introspection, Fragments>
-  : null | _unwrapTypeRec<Type, SelectionSet, Introspection, Fragments>;
+  ? TypeDirective extends 'optional'
```
JoviDeCroock, line 61: Should we add `_optional` here as well or does `_optional extends optional`?

JoviDeCroock (approved):

Issue #10, opened 2024-01-16
RFC: Create a schema builder that can derive a schema type for gql.tada


kitten:
**This is part of a future milestone, and currently proposed to close the client-server gap in GraphQL for small projects.**

## Summary

Currently, GraphQL caters for complex schemas and APIs that are separate from the application they’re built for. However, GraphQL presents a major setup hurdle and makes it hard for people to get started when writing a simple API in a monorepo for a single application, or embedded as an API route into an app.

While this use-case doesn't show all the best strengths of GraphQL at scale and in teams, many individual developers and teams commonly start with one app and one GraphQL schema.

My hypothesis is that it's still too hard to get started with GraphQL compared to [`tRPC`](https://trpc.io/docs/quickstart), which tightly couples server and client-types, and allows for a tight integration between its server-side and client-side code.

While `gql.tada` + `@0no-co/graphqlsp` currently tightens the feedback loop when writing GraphQL code, this can be improved further.

As part of this proposal, `gql.tada` should provide a way to create a small GraphQL schema, which can derive its own schema type in TypeScript.

This schema can then be used directly when authoring GraphQL queries with `gql.tada`, with instant feedback and without reading a schema output from a GraphQL API.

> **Note:** We’re not proposing this to be a replacement for a fully-fledged schema builder, but instead, a schema builder that caters for “embedded” GraphQL API, i.e. single-app GraphQL schemas served on an API route.

## Proposed Solution

`gql.tada/schema` should expose a schema builder, with an API similar to the existing `gqtx`: https://github.com/sikanhe/gqtx

It should provide a helper to derive a dataloader (or other shared-request objects) from the context object on the fly.

It should provide a utility type that either:
- traverses all types on each field recursively, starting from the root types,
- or; a utility type that accepts a schema that accepts an array/object of all defined types

These types should then be converted into [`gql.tada`’s `IntrospectionLikeType`](https://github.com/0no-co/gql.tada/blob/72a4d1b655fcd5ccfaa92721d204823bcdd89ec1/src/introspection.ts#L195-L207) automatically with this utility type and allow a front-end in the same app or monorepo to pass this type to `gql.tada`.

## Requirements

- Should support interfaces, unions, enums, scalars, and object types
- Object types and root type definition shouldn’t be differentiated
- Should output a `GraphQLSchema` and its corresponding definition objects i.e. be a wrapper around graphql.js
- Should provide a type utility to convert a schema to the corresponding `gql.tada` type

[...]

PR #46, opened 2024-01-30
Increase tail-recursive conditional types from 50 to 1000 for variables

### LINKED ISSUE #47: Performance issue in getting enum with huge schema
### Describe the bug

Hello again! It's me with the huge schema generated by Hasura. I'm trying to extract enum type

```ts
type CopayMechanismEnum = ReturnType<typeof graphql.scalar<'plan_copay_mechanisms_enum'>>;
```

Typescript is unable to show hover texts, usage of this type is a no either: assigning any variable to this type won't get any typeahead suggestions either

### Reproduction

https://github.com/deathemperor/gql.tada-reproducible/commit/c746ce8034997bef655c1f7fd59456a2cf2dbd67

### gql.tada version

gql.tada 1.2.0

### Validations

deathemperor:
## Summary

Hello again! As we use gql.tada in our production code, there's an issue arises with variables causing `Type instantiation is excessively deep and possibly infinite`. This is because Hasura's GraphQL variables enable infinite conditions in relationship.

fix #47

## Set of changes

Increase tail-recursive conditional types from 50 to 1000 https://github.com/microsoft/TypeScript/pull/45711

### DIFF
--- a/src/__tests__/fixtures/simpleSchema.ts
+++ b/src/__tests__/fixtures/simpleSchema.ts
@@ -425,4 +425,583 @@ export type simpleSchema = {
       };
     };
   };
+
+  typesList: [
+    {
+      kind: 'INPUT_OBJECT';
+      name: 'TodoPayload';
+      fields: null;
+      inputFields: [
+        {
+          name: 'title';
+          type: {
+            kind: 'NON_NULL';
+            name: null;
+            ofType: {
+              kind: 'SCALAR';
+              name: 'String';
+              ofType: null;
+            };
+          };
+          defaultValue: null;
+        },
+        {
+          name: 'complete';
+          type: {
+            kind: 'SCALAR';
+            name: 'Boolean';
+            ofType: null;
+          };
+          defaultValue: null;
+        },
+      ];
+      interfaces: null;
+      enumValues: null;
+      possibleTypes: null;
+    },
+    {
+      kind: 'OBJECT';
+      name: 'Query';
+      fields: [
+        {
+          name: 'todos';
+          args: [];
+          type: {
+            kind: 'LIST';
+            name: null;
+            ofType: {
+              kind: 'OBJECT';
+              name: 'Todo';
+              ofType: null;
+            };
+          };
+        },
+        {
+          name: 'test';
+          args: [];
+          type: {
+            kind: 'UNION';
+            name: 'Search';
+            ofType: null;
+          };
+        },
+        {
+          name: 'latestTodo';
+          args: [];
+          type: {
+            kind: 'NON_NULL';
+            name: null;
+            ofType: {
+              kind: 'UNION';
+              name: 'LatestTodoResult';
+              ofType: null;
+            };
+          };
+        },
+      ];
+      inputFields: null;
+      interfaces: [];
+      enumValues: null;
+      possibleTypes: null;
+    },
+    {
+      name: 'LatestTodoResult';
+      kind: 'UNION';
+      args: [];
+      possibleTypes: [
+        {
+          kind: 'OBJECT';
+          name: 'Todo';
+          ofType: null;
+        },
+        {
+          kind: 'OBJECT';
+          name: 'NoTodosError';
+          ofType: null;
+        },
+      ];
+    },
+    {
+      kind: 'OBJECT';
+      name: 'NoTodosError';
+      interfaces: [];
+      fields: [
+        {
+          name: 'message';
+          args: [];
+          type: {
+            kind: 'NON_NULL';
+            name: null;
+            ofType: {
+              kind: 'SCALAR';
+              name: 'String';
+              ofType: null;
+            };
+          };
+        },
+      ];
+    },
+    {
+      kind: 'OBJECT';
+      name: 'Todo';
+      fields: [
+        {
+          name: 'id';
+          args: [];
+          type: {
+            kind: 'NON_NULL';
+            name: null;
+            ofType: {
+              kind: 'SCALAR';
+              name: 'ID';
+              ofType: null;
+            };
+          };
+        },
+        {
+          name: 'text';
+          args: [];
+          type: {
+            kind: 'NON_NULL';
+            name: null;
+            ofType: {
+              kind: 'SCALAR';
+              name: 'String';
+              ofType: null;
+            };
+          };
+        },
+        {
+          name: 'complete';
+          args: [];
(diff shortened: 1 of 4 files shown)


JoviDeCroock (changes requested):
The changes to the example aren't in scope of this PR 😅 we don't want your schema/bun to become the norm for our examples. It's also missing a changeset

deathemperor:
@JoviDeCroock my bad, didn't mean to put our schema on the main repo. I've corrected it.

Changing this PR to draft as we've also improve the same issue for fragments and results too. hold on!

src/t.ts
```diff
@@ -0,0 +1,1891 @@
...
+                        getTypeDirective<Node>
+                      >;
+                }
+            : {}) &
+          IntersectionAcc)
+      | UnionAcc
+    >
+  : IntersectionAcc | UnionAcc;
+
+type getPossibleTypeSelectionRec<
```
JoviDeCroock, line 1232: This looks to be the build file can we remove this

deathemperor: thanks, file removed

src/__tests__/fixtures/simpleSchema.ts
```diff
@@ -425,4 +425,583 @@ export type simpleSchema = {
       };
     };
   };
+
+  typesList: [
```
JoviDeCroock, line 429: I don't think `typesList` is a spec compliant property

src/introspection.ts
```diff
@@ -221,11 +221,16 @@ type mapIntrospection<
...
     : never;
   types: mapIntrospectionTypes<Query, Scalars>;
+  introspection: Query;
+  typesList: Query['__schema']['types'];
 };
 
 type getScalarTypeNames<Schema extends IntrospectionLikeType> =
-  Schema['types'][keyof Schema['types']] extends infer Type
-    ? Type extends { kind: 'SCALAR' | 'ENUM'; name: any }
+  Schema['typesList'][number] extends infer Type
```
JoviDeCroock, line 229: typesList does not exist

src/introspection.ts
```diff
@@ -221,11 +221,16 @@ type mapIntrospection<
     ? Query['__schema']['subscriptionType']['name']
     : never;
   types: mapIntrospectionTypes<Query, Scalars>;
+  introspection: Query;
```
JoviDeCroock, line 224: Why are we adding the full introspection here again?

kitten:
I do this rarely, so first I'll apologise 😅 The tl;dr is I'd like to close this PR for now, and address the concerns separately. I'm sorry, and I'm grateful for the time you've put in, but this will need more bisections and validations on our end.

Regarding issue #47 in general, I'm happy to look into the performance of selecting enums.

However, there's also other changes in this PR, all separate and there's no way for me to follow why they've been made and they compromise on clarity in a lot of place:
- The introspection is passed through (duplication)
- A types list is passed on now (duplication)
- Additional generics that I haven't checked yet which aren't quite following the top-down patterns (at best, this requires some refactors, but I also am not sure why they're there)
- There's a separate concern that's being addressed of input objects that are too big (but I don't think the fix lines up with how we're solving this in other places, and may not be effective)

etc; basically, the PR doesn't document what's being addressed, why it's being addressed, and I also can't reproduce all of this.

So, I feel, the best step forward for now here is for me to close this, note down the concerns in #47 and extract the concerns this PR is addressing, and follow along and reproduce this from scratch.
I know this doesn't sound very efficient, but it comes from an approach of “write-twice”, so we're on the same page here and reconcile this with the structure and code we already have two, while also getting a better idea of the performance issues you're seeing.

If this then leads to new PRs — don't worry — we'll include you as a co-author, so you do get credit for the change.

src/selection.ts
```diff
@@ -73,8 +74,15 @@ type getFragmentSelection<
   Type extends ObjectLikeType,
   Introspection extends IntrospectionLikeType,
   Fragments extends { [name: string]: any },
+  UnionAccumulator,
```
kitten, line 77: I don't really know btw what this generic is trying to achieve. It's possible you had a performance concern here, or something else, but the name and placing of this doesn't make it immediately obvious ❤️

src/selection.ts
```diff
@@ -73,8 +74,15 @@ type getFragmentSelection<
...
 > = Node extends { kind: Kind.INLINE_FRAGMENT; selectionSet: any }
-  ? getSelection<Node['selectionSet']['selections'], Type, Introspection, Fragments>
+  ? getSelection<
+      Node['selectionSet']['selections'],
+      Type,
+      Introspection,
+      Fragments,
+      UnionAccumulator
```
kitten, line 84: Ok, I think I know what confused me about this 😅 Sorry!

The `UnionAccumulator` seems to be an attempt to make this tail-recursive. However, there's no direct recursion going on here.
So, it basically adds indirection, since it hoists this result too much.

Just to double-check this, I did benchmark this and the recursion in the other place does work (intersection), since it's isolated to _one_ recursive type, but `UnionAccumulator` doesn't move the benchmarks and doesn't seem to serve a purpose 🤔

kitten:
@deathemperor @HaiNNT: I've extracted the easier tail-recursive fixes into here: https://github.com/0no-co/gql.tada/pull/51
I also found one additional case we missed before (`getVariablesType`)

However, I haven't looked at `gql.scalar` yet. I think, however, I'd want to go with a different approach, but this will take me a bit of time, since I have to get the testing cases for large schemas set up again to reproduce this.

src/introspection.ts
kitten: just to clarify why we're hesitant about `introspection` and `typesList` here. This makes our future plans for #10 complex.

The `mapIntrospection<>` output isn't just an intermediary type to us, but a “format” that we wish to output directly (either in the LSP or in API/schema builders) in the future. So, adding more formats and variations to it will make things difficult.

deathemperor:
thank you @kitten for your very careful and detailed explanation. This PR indeed wasn't good enough to address the concerns it was trying to solve as the DX over the last week was terrible with all the issues this PR caused. And again much appreciated what you've improved in 1.2.1!

However, the `Type instantiation is excessively deep and possibly infinite.ts(2589)` error still exists for type with deep graphql relations (our claim object).

and type checking still costs 5-10s for changes to result in our schema (I've made the reproducible repo private, let me know if you need access). I can also be contacted via Discord deathemperor#7188.

PR #62, opened 2024-02-21
feat: Add `gql.tada/ts-plugin` alias

### LINKED ISSUE #75: RFC: Re-expose the GraphQLSP TS plugin at `gql.tada/ts-plugin`
## Summary

Currently, the [Installation steps](https://gql-tada.0no.co/get-started/installation/#step-1-installing-packages) of `gql.tada` include installing `@0no-co/graphqlsp` and the explanations for how [`gql.tada` functions differentiates between the two packages](https://gql-tada.0no.co/#how-does-it-work).

It may be a small detail, but this often forces us to refer to this project as `gql.tada`, by its package name, in the documentation, and to the TS LSP plugin as `@0no-co/graphqlsp`, which hinders being able to describe how this project functions in short and succinct ways.

Instead, we'd love to be able to automatically instally `@0no-co/graphqlsp` and not expose users to having to understand the internal workings of this structure.

## Proposed Solution

We'd like `@0no-co/graphqlsp` to be re-exported as `gql.tada/lsp` (or a similar import).
This would mean that, in the future, the `tsconfig.json` changes would look as follows:

```json
{
  "compilerOptions": {
    "strict": true,
    "plugins": [
      {
        "name": "gql.tada/lsp",
        "schema": "./schema.graphql",
        "tadaOutputLocation": "./src/graphql-env.d.ts"
      }
    ]
  }
}
```

No additional packages would be installed as `@0no-co/graphqlsp` would become a dependency of `gql.tada` and would be automatically installed.

This previously wasn't possible as TypeScript would block sub-paths from loading as a TS plugin, but will be enabled once this commit is released: https://github.com/microsoft/TypeScript/commit/ef091bf7557f164e9e4294f53941b26a540209c2

## Requirements

- Re-export `@0no-co/graphqlsp` as `gql.tada/lsp` using the existing build pipeline

kitten:
Dependent on https://github.com/microsoft/TypeScript/pull/57266
Resolves #75

## Summary

We're waiting for TypeScript to support sub-paths for tsserver plugins to ease the installation instructions back down to a single package and avoid having to explain the different between `@0no-co/graphqlsp` and `gql.tada` to people.

This would turn `gql.tada` into a standalone entrypoint package for its experience.

## Set of changes

- Add `gql.tada/ts-plugin` alias

### DIFF
--- a/.changeset/big-ravens-drum.md
+++ b/.changeset/big-ravens-drum.md
@@ -1,0 +1,5 @@
+---
+"@gql.tada/cli-utils": minor
+---
+
+Add `gql.tada/ts-plugin` to the init and doctor command
--- a/.changeset/spotty-crabs-report.md
+++ b/.changeset/spotty-crabs-report.md
@@ -1,0 +1,6 @@
+---
+"gql.tada": minor
+"@gql.tada/internal": patch
+---
+
+Add `gql.tada/ts-plugin` alias for `@0no-co/graphqlsp`
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ package-lock.json
 /tsconfig.vitest-temp.json
 /cli
 /internal
+/ts-plugin
--- a/examples/example-pokemon-api/package.json
+++ b/examples/example-pokemon-api/package.json
@@ -13,7 +13,7 @@
     "urql": "^4.0.7"
   },
   "devDependencies": {
-    "@0no-co/graphqlsp": "^1.12.0",
+    "@0no-co/graphqlsp": "^1.12.9",
     "@types/react": "^18.2.79",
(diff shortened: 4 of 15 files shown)


kitten:
Opening as ready since TypeScript `5.5` was announced and a beta (`5.5.0`) is ready. Fix is part of this release: https://github.com/microsoft/TypeScript/commit/ef091bf7557f164e9e4294f53941b26a540209c2

We could consider renaming this to `gql.tada/ts-plugin` instead of `gql.tada/lsp` though.

Another note is that we may want to update the `init` command. We can also add a check for `doctor` against the installed `typescript` version (`<5.5` wouldn't work with this alias of course and we can detect that)

JoviDeCroock (approved):

PR #97, opened 2024-03-02
fix: Allow `graphql.scalar` to resolve types for input objects

kitten:
## Summary

`graphql.scalar()` currently allows scalars and enums to be resolved from the given introspection schema, however, it differed from `getScalarType` in `variables.ts` in that it didn't resolve input objects.

## Set of changes

- Reuse logic in `variables.ts` for `graphql.scalar`

### DIFF
--- a/.changeset/breezy-queens-bow.md
+++ b/.changeset/breezy-queens-bow.md
@@ -1,0 +1,5 @@
+---
+'gql.tada': patch
+---
+
+Allow `graphql.scalar` to resolve types for input objects.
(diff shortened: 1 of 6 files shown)


JoviDeCroock (approved):

llllvvuu:
Is it intentional for all the fields to be optional and nullable?

llllvvuu:
Is it intentional for all of the fields of `ReturnType<typeof graphql.scalar<T>>` to be optional and nullable?

PR #98, opened 2024-03-02
feat: Add `graphql.persisted()` utility to create persisted stand-in documents

### LINKED ISSUE #77: RFC: Implement `graphql.persisted` as a stand-in value for persisted documents
## Summary

Currently, persisted documents are mainly achieved using two methods:
- Either, a hash is created at runtime for documents that are in-source
- or; a compilation step replaces documents with stand-ins with document IDs
In both situations, a “manifest” is often created that contains a mapping of document IDs to GraphQL document strings (`Record<string, string>`) that can then be consumed by server-side GraphQL APIs.

The latter method is often preferred in situations where GraphQL documents _must_ compile away for security and/or obscurity purposes. When only a limited number of GraphQL documents are accepted by the API, compiling them away and keeping them out of the source reduces request payload sizes and keeps the usage of GraphQL completely private.
(This basically turns GraphQL into RPCs)

## Proposed Solution

Currently, `gql.tada` sees being a runtime tool as a benefit, and compilation tools are complex to maintain and complicated tasks, like replacing all `graphql()` calls with shallow ASTs is potentially not as maintainable.

Instead, we're envisioning a stand-in function to be a transparent way to integrate persisted documents.

A stand-in call would look like the following:

```ts
const query = graphql(`
  query HelloQuery {
    hello
    ...HelloWorld
  }
`);

const persistedQuery = graphql.persisted<typeof query>('HelloQuery');
```

The `graphql.persisted` call accepts a GraphQL document as a generic and accepts a document ID as its only argument.

This means that:
- During runtime, `graphql.persisted` only outputs a persisted document — not a full GraphQL document AST
- During compilation, i.e. check-time, this can still carry the type of the full document (`query` in our example above)
- In the compiled JS output, `query` (and all fragments it depends on) compile away, since the calls are side-effect free and are not referred to in JS values (as long as they're not part of other side-effects or assigned to other values)

This has the advantage of leveraging tree-shaking for compilation, and lets users transparently assign document IDs to queries.

### Non-handwritten Document IDs

Since, most commonly, documents use hashes in their IDs, [GraphQLSP](https://github.com/0no-co/graphqlsp) (and a future CLI, See #76) would be tasked to suggest replacing and/or updating the argument with a hash instead.

## Requirements

- The `graphql.persisted` call must transparently take on the type that it's passed by the generic, an act like a full document
- To avoid runtime errors it _may_ take on the shape of `{ kind: 'Document', definitions: [], documentId: string }`, where `definitions` is always an empty array

### External tasks

- We may need a configuration option that tells us the format of document IDs (e.g. `SHA256`)
  - The LSP can then suggest to update/populate the document ID
  - The LSP can allow for a code action to update/populate the document ID, regardless of the configuration option
  - The CLI can lint and highlight outdated document IDs
  - The CLI can auto-fix outdated document IDs, when prompted

Furthermore the CLI **must** allow a manifest to be exported that contains a mapping of document IDs to document strings.

kitten:
Related to #77
Related to https://github.com/0no-co/GraphQLSP/pull/240
Related to https://github.com/urql-graphql/urql/pull/3515

## Summary

This implements a `graphql.persisted()` function, as per the RFC in #77.

This accepts a `documentId` string and returns a value that matches a GraphQL.js `DocumentNode` shape but contains no definitions. The resulting value contains no reference to the original document, while the return type of `graphql.persisted()` matches the document’s type that's been passed in.

The returned type is also transformed to a more strict type that clearly indicates that the `documentId` is present but that no `definitions` are present.

## Set of changes

- Add `graphql.persisted()`
- Add `TadaPersistedDocumentNode`

### DIFF
--- a/.changeset/orange-badgers-decide.md
+++ b/.changeset/orange-badgers-decide.md
@@ -1,0 +1,5 @@
+---
+'gql.tada': minor
+---
+
+Add `graphql.persisted()` to create an API for persisted documents that omits a query’s definitions from the output bundle.
--- a/src/__tests__/api.test-d.ts
+++ b/src/__tests__/api.test-d.ts
@@ -446,3 +446,28 @@ describe('unsafe_readResult', () => {
     expectTypeOf<typeof result>().toEqualTypeOf<ResultOf<document>>();
   });
 });
+
+describe('graphql.persisted()', () => {
+  const graphql = initGraphQLTada<{ introspection: simpleIntrospection }>();
+
+  const query = graphql(`
+    query Test {
+      __typename
+    }
+  `);
+
+  it('should take on the value of the document', () => {
+    const persisted = graphql.persisted<typeof query>('Test');
+    expectTypeOf<typeof persisted>().toMatchTypeOf<typeof query>();
+    expectTypeOf<ResultOf<typeof persisted>>().toEqualTypeOf<ResultOf<typeof query>>();
+    expectTypeOf<VariablesOf<typeof persisted>>().toEqualTypeOf<VariablesOf<typeof query>>();
+  });
+
+  it('should require a document to be passed as a generic', () => {
+    const persisted = graphql.persisted('Test');
+    expectTypeOf<typeof persisted>().toBeNever();
+
+    // @ts-expect-error
+    graphql.persisted<number>('Test');
+  });
+});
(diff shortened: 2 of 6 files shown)


JoviDeCroock (approved):

kitten:
Merging since I expect this to be somewhat usable before the next release

PR #101, opened 2024-03-02
fix: Fix input object field introspection with missing `defaultValue` properties

llllvvuu:
## Summary

If I have no `defaultValue`, `getInputObjectTypeRec` will infer my field as `foo?: string | null` when it should be `foo: string`. This is because `getInputObjectTypeRec` only checks for explicity `undefined | null`. Add a test case and fix for this.

## Set of changes

* Additional "description" field in test fixture, which has no `defaultValue`
* Bugfix and test

### DIFF
--- a/.changeset/strange-nails-ring.md
+++ b/.changeset/strange-nails-ring.md
@@ -1,5 +1,5 @@
 ---
-"@fake-scope/fake-pkg": patch
+'gql.tada': patch
 ---
 
 fix(getInputObjectTypeRec): Correctly infer type of field with no default value (should be non-optional and non-nullable)
--- a/.changeset/strange-nails-ring.md
+++ b/.changeset/strange-nails-ring.md
@@ -2,4 +2,4 @@
 'gql.tada': patch
 ---
 
-fix(getInputObjectTypeRec): Correctly infer type of field with no default value (should be non-optional and non-nullable)
+Handle inference of input object fields with missing `defaultValue` properties in introspection.
--- a/.changeset/strange-nails-ring.md
+++ b/.changeset/strange-nails-ring.md
@@ -1,0 +1,5 @@
+---
+"@fake-scope/fake-pkg": patch
+---
+
+fix(getInputObjectTypeRec): Correctly infer type of field with no default value (should be non-optional and non-nullable)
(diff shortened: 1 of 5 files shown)


.changeset/strange-nails-ring.md
```diff
@@ -0,0 +1,5 @@
+---
+'gql.tada': patch
+---
+
+fix(getInputObjectTypeRec): Correctly infer type of field with no default value (should be non-optional and non-nullable)
```
kitten, line 5:
```suggestion
Fix inference of input object fields with missing `defaultValue` properties in introspection.
```

src/variables.ts
```diff
@@ -12,9 +12,20 @@ type getInputObjectTypeRec<
...
+              : {
+                  [Name in InputField['name']]?: unwrapType<
+                    InputField['type'],
+                    Introspection
+                  > | null;
+                }
+            : { [Name in InputField['name']]: unwrapType<InputField['type'], Introspection> }
+          : {
+              [Name in InputField['name']]?: unwrapType<InputField['type'], Introspection> | null;
+            }
```
kitten, line 28:
```suggestion
        ? InputField extends { defaultValue?: undefined | null; type: { kind: 'NON_NULL' } }
          ? { [Name in InputField['name']]: unwrapType<InputField['type'], Introspection> }
          : { [Name in InputField['name']]?: unwrapType<InputField['type'], Introspection> | null }
```

This is a bit of a wordy change and complexity increase in the types.
What about this change? It seems to basically be able to do the same thing and be equivalent.

kitten (commented):
Nice catch!

I've got a suggestion for the changeset (just a note, this is the exact text that goes in the changelog, so it needs to be user-facing). Same suggestion is already applied to the PR title (`fix(x)` is generally reserve for sub-packages)

I've got a larger suggestion here for the type change. It seems a bit long and wordy, and as far as I can tell, is a complex multi-part match for a change we can probably do by just adding `?` to `defaultValue`.
Since it's typed as `defaultValue?: T | null`, and `T | undefined | null` isn't equivalent, this minimal change should do the trick and seems to pass tests for me locally.

llllvvuu:
Thanks for the suggestions! I didn't know about `?` and it works for me.

PR #118, opened 2024-03-07
docs: Fix customization instructions

nathanchapman:
## Summary

Fix sample code in customization instructions on website for these two sections:
- Initializing gql.tada manually
- Customizing scalar types

### DIFF
--- a/website/src/content/docs/get-started/installation.mdx
+++ b/website/src/content/docs/get-started/installation.mdx
@@ -296,10 +296,10 @@ is a string.
 
 ```ts {"Define scalar types here:":6-10} title="src/graphql.ts"
 import { initGraphQLTada } from 'gql.tada';
-import type { introspection } from './graphql-env.d.ts';
+import { introspection } from './graphql-env.d.ts';
 
 export const graphql = initGraphQLTada<{
-  introspection: introspection;
+  introspection: typeof introspection;
 
   scalars: {
     DateTime: string,
--- a/website/src/content/docs/get-started/installation.mdx
+++ b/website/src/content/docs/get-started/installation.mdx
@@ -266,7 +266,7 @@ To work around this, we’ll create a file that uses the introspection data manu
 
 ```ts title="src/graphql.ts" {4-6}
 import { initGraphQLTada } from 'gql.tada';
-import { introspection } from './graphql-env.d.ts';
+import { introspection } from './graphql-env.ts';
 
 export const graphql = initGraphQLTada<{
   introspection: typeof introspection;
@@ -296,7 +296,7 @@ is a string.
 
 ```ts {"Define scalar types here:":6-10} title="src/graphql.ts"
 import { initGraphQLTada } from 'gql.tada';
-import { introspection } from './graphql-env.d.ts';
+import { introspection } from './graphql-env.ts';
 
 export const graphql = initGraphQLTada<{
   introspection: typeof introspection;
--- a/website/src/content/docs/get-started/installation.mdx
+++ b/website/src/content/docs/get-started/installation.mdx
@@ -266,10 +266,10 @@ To work around this, we’ll create a file that uses the introspection data manu
 
 ```ts title="src/graphql.ts" {4-6}
 import { initGraphQLTada } from 'gql.tada';
-import type { introspection } from './graphql-env.d.ts';
+import { introspection } from './graphql-env.d.ts';
 
 export const graphql = initGraphQLTada<{
-  introspection: introspection;
+  introspection: typeof introspection;
 }>();
 
 export type { FragmentOf, ResultOf, VariablesOf } from 'gql.tada';


kitten:
This seems out-of-date to me, as `@0no-co/graphqlsp` now outputs a different format. Basically it adds an `export type` there, so the `d.ts` output outputs a `type`, while the `.ts` output outpus a value.
Did you make sure to try this out with the latest version's output? ✌️ 
(You can now also output it with the CLI)

nathanchapman:
Hey @kitten! Yes, this is against the latest version of `@0no-co/graphqlsp` and `gql.tada`
Good point about the `.d.ts` vs `.ts` files. These instructions are for customizing, so the examples should be using a `.ts` file

Instructions/comment from a `.d.ts` file:
```ts
/** An IntrospectionQuery representation of your schema.
 *
 * @remarks
 * This is an introspection of your schema saved as a file by GraphQLSP.
 * It will automatically be used by `gql.tada` to infer the types of your GraphQL documents.
 * If you need to reuse this data or update your `scalars`, update `tadaOutputLocation` to
 * instead save to a .ts instead of a .d.ts file.
 */
```

I updated the examples in this PR to import from a `.ts` file as well

kitten:
That doesn't seem quite right though, sorry 😅 ❤️ 

The important detail here is that the docs explicitly recommend saving the introspection to a `.d.ts` file, which outputs the introspection as a type. So, your change explicitly only works if a user was to switch to a `.ts` output, which isn't mentioned in the docs at that point yet.

Instead, the intention is that this is only mentioned once you get to this file: https://gql-tada.0no.co/reference/graphqlsp-config/#tadaoutputlocation

That's a change we had to make because:
- the `.d.ts` format is a lot more efficient for TS to parse and process, and `export type` simply makes more sense there
- the `.ts` format was already used by some folks at that point and the intention there is that it's usable by value (however, the original instructions did show `typeof`)
- the `.d.ts` format is likely going to diverge further so less work is done by the type inference in `introspection.ts`, but that's a little besides the point

Hope that makes sense ❤️

nathanchapman:
Hmm.. two questions:

One of the two sections I updated is titled `Customizing scalar types`
How would someone go about customizing scalars etc. with the `.d.ts` file?

I'm also working with a multi-schema monorepo (3 applications interacting with 3 different GraphQL APIs)
This type of setup is mentioned in the other section I updated here titled `Initializing gql.tada manually`
How would someone achieve this with a `.d.ts` file?

kitten:
> I'm also working with a multi-schema monorepo 

This isn't supported yet 🥲 It's unfortunately a little tricky. That's because `GraphQLSP` cannot differentiate between schemas and doesn't know which schema to check against for any given document.

We have an idea for addressing this, but that may take us a bit of time.

The way to address this for now is to instantiate `gql.tada` multiple times, i.e. passing different schemas to `initGraphQLTada<>()`. It doesn't matter where the schema lives, so that means you can write multiple `graphql-env.d.ts` files, e.g. `schema-1.d.ts`, and so on.

However, currently, there's no mechanism to generate these separately, unless you were to isolate the instances of `gql.tada` in faux-packages/monorepo packages with their own `tsconfig.json`s and used either GraphQLSP just in that package or by using the CLI there.

In short, multi-schema use-cases aren't supported yet.
I'm very eager to make these happen though. While we can talk a lot about how ideally an organisation shouldn't have multiple schemas they're consuming, since those often cross boundaries and this opens you up to having apps that also consume multiple schemas (which starts defeating the point of GraphQL pretty hard),
I ran into this myself in a simple case too with a monorepo that both had a GraphQL schema but also consumed a third-party GraphQL schema.

Anyway, that's a little off topic ❤️ Maybe better suited for Discord (where this question was also asked, I believe)
Closing for now

nathanchapman:
> This isn't supported yet 🥲 It's unfortunately a little tricky. That's because GraphQLSP cannot differentiate between schemas and doesn't know which schema to check against for any given document.
>
> We have an idea for addressing this, but that may take us a bit of time.
>
> The way to address this for now is to instantiate gql.tada multiple times, i.e. passing different schemas to initGraphQLTada<>(). It doesn't matter where the schema lives, so that means you can write multiple graphql-env.d.ts files, e.g. schema-1.d.ts, and so on.
>
> However, currently, there's no mechanism to generate these separately, unless you were to isolate the instances of gql.tada in faux-packages/monorepo packages with their own tsconfig.jsons and used either GraphQLSP just in that package or by using the CLI there.
>
> In short, multi-schema use-cases aren't supported yet. I'm very eager to make these happen though. 

It's working just fine for me.. but again, **not** with `.d.ts` file(s) since they'd compete with one another for the __only__ introspection value:

**tsconfig.json**
```json
{
  "compilerOptions": {
    "plugins": [
        {
          "name": "@0no-co/graphqlsp",
          "schema": "./src/app1/schema.graphql",
          "tadaOutputLocation": "./src/app1/tada.ts"
        },
        {
          "name": "@0no-co/graphqlsp",
          "schema": "./src/app2/schema.graphql",
          "tadaOutputLocation": "./src/app2/tada.ts"
        },
        {
          "name": "@0no-co/graphqlsp",
          "schema": "./src/app3/schema.graphql",
          "tadaOutputLocation": "./src/app3/tada.ts"
        }
      ],
  }
}
```

**example `tada.ts` (e.g. app1, but all 3 would look identical)**
```ts
import { initGraphQLTada } from 'gql.tada'
import { introspection } from './tada'
import { type Scalars } from 'src/shared/scalars'

export const graphql = initGraphQLTada<{
  introspection: typeof introspection
  scalars: Scalars
}>()

export type { FragmentOf, ResultOf, VariablesOf } from 'gql.tada'
export { readFragment } from 'gql.tada'
```

> While we can talk a lot about how ideally an organisation shouldn't have multiple schemas they're consuming, since those often cross boundaries and this opens you up to having apps that also consume multiple schemas (which starts defeating the point of GraphQL pretty hard),
I ran into this myself in a simple case too with a monorepo that both had a GraphQL schema but also consumed a third-party GraphQL schema.

It's fairly standard to have distinct GraphQL APIs for distinct applications (even when they're all "first party"). They have different use cases, capabilities, auth patterns, etc.

My point in opening up this PR was to help other people get their setups working. We kinda glossed over the fact that "monorepo" was mentioned in section I was trying to update the documentation for and that `.d.ts` definitely isn't working with a multi-schema monorepo setup at the moment, but using `.ts` files does in fact work as expected

PR #120, opened 2024-03-08
feat(cli): add generate schema command

JoviDeCroock:
## Summary

Add `generate-schema` command which takes a URL | path to a JSON file and outputs a graphql schema. Example:

- `gql-tada generate-schema https://example.com --header 'authorization: bearer token'`
- `gql-tada generate-schema ./introspection.json`

### DIFF
--- a/packages/cli-utils/src/index.ts
+++ b/packages/cli-utils/src/index.ts
@@ -18,14 +18,18 @@ prog.version(process.env.npm_package_version || '0.0.0');
 
 async function main() {
   prog
-    .command('generate-schema <target> <dest>')
+    .command('generate-schema <target>')
     .describe(
       'Generate a GraphQL schema from a URL or introspection file, this will be generated from the parameters to this command.'
     )
     .option('--header', 'Pass a header to be used when fetching the introspection.')
+    .option(
+      '--output',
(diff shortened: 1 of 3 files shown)


packages/cli-utils/src/index.ts
```diff
@@ -1,58 +1,101 @@
...
+async function main() {
+  prog
+    .command('generate-schema <target> <dest>')
+    .describe(
+      'Generate a GraphQL schema from a URL or introspection file, this will be generated from the parameters to this command.'
+    )
+    .option('--header', 'Pass a header to be used when fetching the introspection.')
+    .example("generate-schema https://example.com --header 'Authorization: Bearer token'")
+    .example('generate-schema ./introspection.json')
+    .action(async (target, destination, options) => {
```
kitten, line 28: It probably makes sense to read the destination from the `tsconfig.json` (similar to: https://github.com/0no-co/gql.tada/blob/b09e2649ec9e6fd57b4290365ec1e1951058e12c/packages/cli-utils/src/index.ts#L86)

Both commands could later on allow for an override option (e.g. `-o --output`, I guess. But we have the benefit of knowing where that schema should end up

kitten (approved):

PR #135, opened 2024-03-13
feat(cli): export cli actions as node apis

### LINKED ISSUE #134: RFC: Exposing the `generate-schema` and `generate-output` actions from `cli-utils` as Node.js API's
## Summary

We maintain a CLI to help users bootstrap a new project on our platform (e.g., `npm create ...`), and as part of that process we would like to:

1. Run through our normal bootstrap process
2. Generate a GraphQL schema
3. Output the introspection file for `gql.tada` to use

While we could achieve this by using `child_process.exec` to call the `gql.tada` CLI from within our CLI, we have concerns around:

- Type safety: We prefer strongly typed method calls over string-based CLI commands
- Performance: Using `exec` spawns a new shell which adds overhead, but a direct API call can be executed within the same process space as our CLI
- Error handling: Parsing `stdout` and `stderr` from `exec` to determine success/failure can become complicated

## Proposed Solution

1. Determine a reasonable API interface for the methods to be exposed (e.g., making `cwd` an optional parameter on the method)
2. Abstracting `generate-schema` and `generate-output` logic into two individual methods that live outside of the CLI action handler callbacks.
3. Exporting those methods as named exports from the `@gql.tada/cli-utils` package (we think importing these methods from `@gql.tada/cli-utils` makes the most sense so we don't have to install the entire `gql.tada` package in our CLI)

## Requirements

N/A

## Relevant Links

- More information on the CLI in reference: #76

matthewvolk:
## Summary

Resolves #134

Abstract core logic for `generate-schema` and `generate-output` CLI commands into importable Node.js APIs.

## Set of changes

The diff is large, but the code is largely the same, mostly just moved around:

- All of the logic in `generateSchema` is unchanged, with the exception of header parsing being pulled out of the function and now headers are passed in as a `Record<string,string>`. Additionally, `cwd` is now an optional parameter (helpful when this methods are called from a directory outside of the directory containing your project's `tsconfig.json` file, e.g., `npm create some-project`).
- `generateTadaTypes` is almost entirely unchanged, `cwd` is now an optional parameter.

### DIFF
--- a/packages/cli-utils/src/index.ts
+++ b/packages/cli-utils/src/index.ts
@@ -12,6 +12,140 @@ import type { GraphQLSPConfig } from './lsp';
 import { hasGraphQLSP } from './lsp';
 import { ensureTadaIntrospection } from './tada';
 
+interface GenerateSchemaOptions {
+  headers?: Record<string, string>;
+  output?: string;
+  cwd?: string;
+}
+
+export async function generateSchema(
+  target: string,
+  { headers, output, cwd = process.cwd() }: GenerateSchemaOptions
+) {
+  let url: URL | undefined;
+
+  try {
+    url = new URL(target);
+  } catch (e) {}
+
+  let introspection: IntrospectionQuery;
+  if (url) {
+    const response = await fetch(url!.toString(), {
+      method: 'POST',
+      headers: {
+        ...headers,
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify({
+        query: getIntrospectionQuery({
+          descriptions: true,
+          schemaDescription: false,
+          inputValueDeprecation: false,
+          directiveIsRepeatable: false,
+          specifiedByUrl: false,
+        }),
+      }),
+    });
+
+    if (response.ok) {
+      const text = await response.text();
+
+      try {
+        const result = JSON.parse(text);
+        if (result.data) {
+          introspection = (result as { data: IntrospectionQuery }).data;
+        } else {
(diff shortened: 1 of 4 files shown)


JoviDeCroock (approved):

JoviDeCroock:
Also needs a changeset

packages/cli-utils/src/index.ts
```diff
@@ -12,11 +12,145 @@ import type { GraphQLSPConfig } from './lsp';
...
+    if (!foundPlugin.schema.endsWith('.graphql')) {
+      console.error(`Found "${foundPlugin.schema}" which is not a path to a GraphQL Schema.`);
+      return;
+    }
+  }
+
+  await fs.writeFile(resolve(cwd, destination), printSchema(schema), 'utf-8');
+}
+
+export async function generateOutput(cwd: string = process.cwd()) {
```
JoviDeCroock, line 121: Let's rename this to `generateTadaTypes` or something that matches the CLI command more

src/cli/index.ts
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env node
 
 import cli from '@gql.tada/cli-utils';
-cli();
+(cli as any).default();
```
JoviDeCroock, line 4: Are the types erroneously generated now?

matthewvolk: Confirmed types for `gql-data-cli.d.ts` are the same as they were previous to this change 👍🙂

PR #146, opened 2024-03-20
fix: Update default ID scalar type to be `string`

### LINKED ISSUE #145: `ID` type should be `string` not `string | number`
### Describe the bug

gql.tada specifies the default scalar `ID` as `string | number`:

https://github.com/0no-co/gql.tada/blob/e8820360ae29af899dc1b922c55839bbf4d3eaf2/src/introspection.ts#L112-L113

From [the GraphQL Specification, 3.5.5 `ID`](http://spec.graphql.org/draft/#sec-ID) (emphasis mine):

> The ID scalar type represents a unique identifier, often used to refetch an object or as the key for a cache. The ID type is **serialized in the same way as a `String`**; however, it is not intended to be human-readable. While it is often numeric, **it should always serialize as a `String`**.

I think this `should` should have been `must`, but the main point is an `ID` type should never come back from the server with `typeof id === 'number'`, it will always be `typeof id === 'string'` (or `null`).

### Reproduction

_No response_

### gql.tada version

`main` branch

### Validations

JoviDeCroock:
Resolves #145 

## Summary

In the spec it is outlined that `ID` should always serialize to `string` https://spec.graphql.org/October2021/#sec-ID

### DIFF
--- a/.changeset/two-rockets-hear.md
+++ b/.changeset/two-rockets-hear.md
@@ -2,4 +2,4 @@
 "gql.tada": minor
 ---
 
-Change the default `scalar` type of `ID` to be `string`, as specified in [the spec](https://spec.graphql.org/October2021/#sec-ID)
+Change the default scalar type of `ID` to be `string`, as [the GraphQL spec](https://spec.graphql.org/October2021/#sec-ID) recommends it to serialize to a string
--- a/.changeset/two-rockets-hear.md
+++ b/.changeset/two-rockets-hear.md
@@ -1,0 +1,5 @@
+---
+"gql.tada": minor
+---
+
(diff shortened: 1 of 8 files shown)


.changeset/two-rockets-hear.md
```diff
@@ -0,0 +1,5 @@
+---
+"gql.tada": minor
+---
+
+Change the default `scalar` type of `ID` to be `string`, as specified in [the spec](https://spec.graphql.org/October2021/#sec-ID)
```
kitten, line 5:
```suggestion
Change the default scalar type of `ID` to be `string`, as [the GraphQL spec](https://spec.graphql.org/October2021/#sec-ID) recommends it to serialize to a string
```

abdonrd:
It would be nice if you could update the readme image.
It shows `id: string | number;`

https://github.com/0no-co/gql.tada?tab=readme-ov-file#-lets-take-a-look

PR #147, opened 2024-03-20
refactor: Accept pre-processed schemas as introspection inputs

kitten:
## Summary

This allows the API to accept a pre-processed “`gql.tada` schema” to be accepted as the `introspection` option (i.e. on `AbstractSetupSchema` / `setupSchema`). Effectively, this means that `mapIntrospection` is used on demand, and only when the input hasn't already been run through `mapIntrospection` before.

This also required us to then split introspection mapping into two steps — one for the general remapping, and one for scalars.

## Set of changes

- Split `mapIntrospection` into non-scalar and scalar-adding steps
- Accept pre-processed schema as `introspection` option

### DIFF
--- a/src/introspection.ts
+++ b/src/introspection.ts
@@ -117,19 +117,6 @@ interface DefaultScalars {
   Int: number;
 }
 
-type mapScalar<
-  Type extends IntrospectionScalarType,
-  Scalars extends ScalarsLike = DefaultScalars,
-> = {
-  kind: 'SCALAR';
-  name: Type['name'];
-  type: Type['name'] extends keyof Scalars
-    ? Scalars[Type['name']]
-    : Type['name'] extends keyof DefaultScalars
-      ? DefaultScalars[Type['name']]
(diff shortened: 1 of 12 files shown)


src/introspection.ts
```diff
@@ -181,57 +168,77 @@ type mapUnion<T extends IntrospectionUnionType> = {
...
-  Query extends IntrospectionQuery,
+type mapIntrospectionScalarTypes<Scalars extends ScalarsLike = DefaultScalars> = obj<{
+  [P in keyof Scalars | keyof DefaultScalars]: {
+    kind: 'SCALAR';
+    name: P;
+    type: P extends keyof Scalars
+      ? Scalars[P]
+      : P extends keyof DefaultScalars
+        ? DefaultScalars[P]
+        : 'wat';
```
JoviDeCroock, line 201:
```suggestion
        : never;
```

JoviDeCroock (approved):

PR #158, opened 2024-03-21
docs: hasura integration

JoviDeCroock:
## Summary

This details how you can shrink your Hasura schema for GraphQLSP


anirudhsama:
When does the `.d.ts` file get regenerated? Say I make a change in the hasura console, do I have to restart the typescript server so that the new file with updated nodes get created?

JoviDeCroock:
@anirudhsama Depends on how you are polling your schema... If you leverage a `schema.graphql` it will watch the file, if you leverage HTTP based introspection it will re-poll every minute or so

anirudhsama:
Thanks for the response. We are using Hasura so I'm using the http based introspection. One limitation right now is that there is no way to pass the secret via an environment variable since the config is in tsconfig.json right?

JoviDeCroock:
No, but you can use the `gql.tada` CLI for that

anirudhsama:
Ok so we can omit the schema and tadaOutputFile in the config if it’s generated via the CLI?

kitten:
@anirudhsama: Please use a separate GitHub discussion or issues for questions, if it doesn't relate to the main PR/issue you're replying to ❤️ 

---

> Ok so we can omit the schema and tadaOutputFile in the config if it’s generated via the CLI?

No, if you can only pass the header in certain, limited cases, you can use `$ gql.tada generate-schema` to pull the schema, which has no watch mode though, and then use the resulting `schema.graphql` file in your `tsconfig`.

JoviDeCroock:
Closing due to inactivity, can always be picked up again

PR #165, opened 2024-03-28
fix(internal): Improve loader by adding a merged loader and incorporating `SchemaOrigin` parsing

kitten:
## Summary

This merges the `SchemaOrigin` type into `@gql.tada/internal` and merges both loaders into one. It also changes the loader interface to automatically reload schemas & introspections on updates and to add a single method to load both the schema and the introspection.

The latter methods are now marked as internal and won't reload the internal data.

## Set of changes

- Add `.load()` to loaders`
- Mark old loading methods as internal
- Reload schema/introspection data when watcher notifies
- Add merged `load()` function that accepts `SchemaOrigin`

### DIFF
--- a/packages/cli-utils/src/tada.ts
+++ b/packages/cli-utils/src/tada.ts
@@ -4,13 +4,11 @@ import type { IntrospectionQuery } from 'graphql';
 import type { SchemaLoader } from '@gql.tada/internal';
 
 import {
+  type SchemaOrigin,
   minifyIntrospection,
   outputIntrospectionFile,
-  loadFromSDL,
-  loadFromURL,
+  load,
 } from '@gql.tada/internal';
-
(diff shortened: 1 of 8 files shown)


packages/cli-utils/src/tada.ts
```diff
@@ -53,27 +51,6 @@ export async function ensureTadaIntrospection(
...
-    } catch (error) {
-      throw new Error(`Input URL "${origin.url}" is invalid`);
-    }
-  } else {
-    return null;
-  }
-};
-
 export function makeLoader(root: string, origin: SchemaOrigin): SchemaLoader {
   const urlOrigin = getURLConfig(origin);
```
JoviDeCroock, line 55: Some issues

JoviDeCroock (approved):

Issue #76, opened 2024-02-24
RFC: Implement a CLI for programmatic TS LSP & Checking tasks


kitten:
## Summary

Currently we have a split in what the TS LSP plugin validates and what `gql.tada` validates in TS typings.

<table>
<thead><tr>
<td><strong><code>gql.tada</code></strong></td>
<td><strong><code>@0no-co/graphqlsp</code></strong></td>
</tr></thead>
<tr><td>

- Errors on field and sub-selection types, if they don't match usage
- Errors, potentially, on invalid fragments and fragment spreads

</td><td>

- Errors on unused fields and fragments
- Errors on unknown fragments
- Errors on invalid types passed to field arguments
- (other schema-side validation)

</td></tr>
</table>

This creates a split experience, where the in-editor experience is cohesive, prompt, and accurate, but the `tsc` (i.e. type checking) step when **compilation at build-time** occurs lacks everything in `@0no-co/graphqlsp`, i.e. the TS LSP plugin.

This can further be summarised by the following issues:
- Set up isn't automated
- Checking whether the setup is correct isn't possible
- Linting isn't possible at build-time
- Static analysis (e.g. for persisted queries) isn't possible

## Proposed Solution

We'd like to collect tasks for a CLI and implement them, possibly gradually over time.
A CLI would be runnable either as `$ gql.tada` or `$ gql-tada`, with the following sub-commands:

- [x] `gql-tada init`: For setting up a project
- [x] `gql-tada doctor`: For checking pre-condition (TypeScript version, `tsconfig.json`, etc)
  - This may also make suggestions re. the setup to the user that aren't critical to `gql.tada` functioning properly (e.g. for performance reasons)
- [x] `gql-tada check`: For checking/linting steps in the TS LSP (document validation, unused fields, etc) as lint rules
  - This may allow for auto-fixing rules, similar to what `eslint --fix` does
- [x] `gql-tada generate-output`: For generating `tadaOutputLocation` i.e. generated file
- [x] `gql-tada generate-schema`: For generating a `schema` that the TS LSP plugin can consume
  - This is a pain point for projects where no other tool outputs a `schema.graphql` file and can be supplemented by this CLI
- [x] `gql-tada generate-persisted`: For generating a list of persisted documents (**TBD** for future RFC)

The exact names of these subcommands isn't determined yet.

## Requirements

**TBD**

- `gql-tada check` and `gql-tada generate-output` must be runnable in CI

wyattades:
[...]

PR #169, opened 2024-04-03
feat(cli): doctor command

JoviDeCroock:
## Summary

Addresses part of https://github.com/0no-co/gql.tada/issues/76

This adds `gql.tada doctor` which does the following

- Check whether we have a `package.json`
- Check whether we have a `typescript` dependency > 4.1.0 (when string literal parsing got introduced)
- Check whether `gql.tada` is intalled
- Check whether `@0no-co/graphqlsp` is installed
- Check whether we have a `tsconfig`
- Check whether GraphQLSP is in the tsconfig
- Check whether the `schema` setting is valid
- Check whether the `tadaOutputLocation` is present

### DIFF
--- a/packages/cli-utils/package.json
+++ b/packages/cli-utils/package.json
@@ -49,7 +49,8 @@
   },
   "dependencies": {
     "@gql.tada/internal": "workspace:*",
-    "graphql": "^16.8.1"
+    "graphql": "^16.8.1",
+    "semiver": "^1.1.0"
   },
   "peerDependencies": {
     "typescript": "^5.0.0"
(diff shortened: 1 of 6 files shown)


packages/cli-utils/package.json
```diff
@@ -49,7 +49,8 @@
   },
   "dependencies": {
     "@gql.tada/internal": "workspace:*",
-    "graphql": "^16.8.1"
+    "graphql": "^16.8.1",
+    "semiver": "^1.1.0"
```
kitten, line 53: Let's move this to `devDependencies` and bundle it away, for the same reasons as we bundle most other things away in `cli-utils`?

kitten (approved):

PR #170, opened 2024-04-04
feat(cli): check command

JoviDeCroock:
## Summary

Addresses part of https://github.com/0no-co/gql.tada/issues/76

This adds `gql.tada check --severity error` which gets the current TS Project and runs the diagnostics from the LSP project


packages/internal/src/diagnostics/index.ts
```diff
@@ -0,0 +1,137 @@
...
+
+type GraphQLSPConfig = {
+  name: string;
+  schema: SchemaOrigin;
+  tadaOutputLocation: string;
+};
+
+export async function check(
+  config: GraphQLSPConfig,
+  minSeverity: Severity = 'error'
```
kitten, line 27: (doesn't block merging, can be a todo)

I'd argue this shouldn't be here and should be handled by the CLI. This is otherwise a little ambiguous.

At the CLI level, we could have a `--log-level/--level` option that basically just hides verbose output when it's below the specified level

packages/cli-utils/src/index.ts
```diff
@@ -155,8 +138,95 @@ async function main() {
...
+  }
+
+  const config = getGraphQLSPConfig(tsConfig);
+  if (!config) {
+    console.error(`Could not find a "@0no-co/graphqlsp" plugin in your tsconfig.`);
+    return;
+  }
+
+  return config;
+};
```
kitten, line 230: (doesn't block merging, can be a todo)

We'll have to start pulling the actions out into separate files. 

On a side-note, I'm not a huge fan of `sade` because it doesn't encourage splitting up sub-commands and the main commands (i.e. nested command levels with separate options being co-located with the action)

But anyway, main point, let's already start pulling action functions out

**Strong suggestion to not do this in this PR to not waste time merging this**

packages/cli-utils/src/index.ts
```diff
@@ -155,8 +138,95 @@ async function main() {
...
+        const warningsReport = warnDiagnostics.length
+          ? `Found ${warnDiagnostics.length} Warnings:\n\n${constructDiagnosticsPerFile(
+              warnDiagnostics
+            )}\n\n`
+          : ``;
+        const suggestionsReport = infoDiagnostics.length
+          ? `Found ${infoDiagnostics.length} Suggestions:\n\n${constructDiagnosticsPerFile(
+              infoDiagnostics
+            )}\n\n`
+          : ``;
```
kitten, line 176: (doesn't block merging, can be a todo)

I think, the aforementioned `--level` could filter this, rather than that being in the external function.

On another note, eslint does something clever here.
They'd hide `info`-level diagnostics when there's nothing that breaks the output.

i.e. when there's no error/warning, we wouldn't output `info` and only add a filter

packages/cli-utils/src/index.ts
```diff
@@ -155,8 +138,95 @@ async function main() {
...
+        // eslint-disable-next-line no-console
+        console.log(`${errorReport}${warningsReport}${suggestionsReport}`);
```
kitten, line 178: (doesn't block merging, can be a todo)

We'd need a summary to separate the error/warn/info list from their respective counts

There's also no `process.exit` codes here yet

packages/cli-utils/src/index.ts
```diff
@@ -155,8 +138,95 @@ async function main() {
       const shouldPreprocess =
         !options['disable-preprocessing'] && options['disable-preprocessing'] !== 'false';
       return generateTadaTypes(shouldPreprocess);
+    })
+    .command('check')
+    .option('--severity', 'The minimum severity of diagnostics to display.')
```
kitten, line 143: (doesn't block merging, can be a todo)

As per the other comment, I'd suggest a `--level` but let's maybe also add a `--skip-warnings` or `--fail-on-warn` and by default not fail on warnings (and similarly to info, hide detailed outputs when we succeed / exit with code `0`)

packages/internal/src/diagnostics/index.ts
```diff
@@ -0,0 +1,137 @@
+import { Project, ts } from 'ts-morph';
```
kitten, line 1: re. `internal/src/diagnostics/index.ts`

Starting to think about this, do we want this to be in `@gql.tada/internal`? I don't really see how that'd be beneficial because it's not unlikely that we'll split `check` up further (since it currently does _a lot_ e.g. instantiate the project, call `resolveTypeScriptRootDir`, call `load`)

It feels more like a distinct file to add to the CLI. If we add it to `@gql.tada/internal` I'd expect it to be used by GraphQLSP (or other future tools), but I'm not sure if it ever will be in the current shape 😅

kitten (approved):
I don't think anything's strictly blocking the merge, assuming we'll keep iterating, so nothing here blocks this and we're good to go on the main implementation

packages/internal/src/diagnostics/index.ts
```diff
@@ -0,0 +1,137 @@
...
+
+export async function check(
+  config: GraphQLSPConfig,
+  minSeverity: Severity = 'error'
+): Promise<FormattedDisplayableDiagnostic[]> {
+  const projectName = path.resolve(process.cwd(), 'tsconfig.json');
+
+  const project = new Project({
+    tsConfigFilePath: './tsconfig.json',
+  });
```
kitten, line 33: Question (not a blocker): I'm a bit concerned about this being hardcoded to `./tsconfig.json`. And the same goes for other parts of the CLI using mainly `${CWD}/tsconfig.json`.

I tested `new Project()` and it seems to indeed not trigger the resolver, so todo here to figure out resolver, which could be grouped with the polishing task to figure out reusing `TsConfigResolver` from `@ts-morph/common`, and then to also get to replacing `resolveTypeScriptRootDir` with something externally maintained, as discussed 🤔 ❤️

packages/internal/src/diagnostics/index.ts
```diff
@@ -0,0 +1,137 @@
...
+      },
+      // This prevents us from exposing normal diagnostics
+      getSemanticDiagnostics: () => [],
+    } as any,
+    languageServiceHost: {} as any,
+    project: {
+      getProjectName: () => path.resolve(process.cwd(), 'tsconfig.json'),
+      readFile: (p) => {
+        return fs.readFileSync(p);
+      },
```
kitten, line 69: TODO for myself: I'll need to look at this closer. This seems a bit odd, since I assume, this is somehow fed from the `Host` accessing FS. In this case, we may have some unintended blocking FS ops here

packages/cli-utils/src/index.ts
JoviDeCroock: Added a process.exit, no summary yet as I figured we have no final thoughts on the view yet.

packages/internal/src/diagnostics/index.ts
JoviDeCroock: Moved it

examples/example-pokemon-api/package.json
```diff
@@ -10,6 +10,7 @@
     "gql.tada": "^1.4.0",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
+    "ts-morph": "^22.0.0",
```
kitten, line 13: Accidental addition?

```suggestion
```

packages/cli-utils/package.json
```diff
@@ -49,6 +49,8 @@
   },
   "dependencies": {
     "@gql.tada/internal": "workspace:*",
+    "@0no-co/graphqlsp": "^1.8.0",
+    "ts-morph": "^22.0.0",
```
kitten, line 53:
```suggestion
    "ts-morph": "~22.0.0",
```

Should we make this more strict?

kitten (approved):

PR #176, opened 2024-04-05
feat(cli): generate persisted operations json files

JoviDeCroock:
## Summary

This adds a new command called `generate-persisted` which will go through all the files and generate a JSON file containing all your persisted operations. This can then be used to upload to GraphQL Yoga or some other server to limit the distinct amount of documents that can be sent to the GraphQL server.

Related to https://github.com/0no-co/GraphQLSP/pull/240
Implements part of #76

### DIFF
--- a/packages/cli-utils/src/commands/generate-persisted.ts
+++ b/packages/cli-utils/src/commands/generate-persisted.ts
@@ -62,25 +62,35 @@ async function getPersistedOperationsFromFiles(
 
   return sourceFiles.reduce((acc, sourceFile) => {
     const persistedCallExpressions = findAllPersistedCallExpressions(sourceFile.compilerNode);
+    const currentFilename = sourceFile.compilerNode.fileName;
     return {
       ...acc,
       ...persistedCallExpressions.reduce((acc, callExpression) => {
         const hash = callExpression.arguments[0].getText();
         if (!callExpression.typeArguments) {
+          console.warn(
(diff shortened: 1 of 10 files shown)


packages/cli-utils/src/commands/generate-persisted.ts
```diff
@@ -0,0 +1,143 @@
...
+          return acc;
+        }
+
+        const fragments = [];
+        const operation = initializer.arguments[0].getText().slice(1, -1);
+        if (initializer.arguments[1] && ts.isArrayLiteralExpression(initializer.arguments[1])) {
+          unrollTadaFragments(initializer.arguments[1], fragments, pluginCreateInfo);
+        }
+
+        const document = `${operation}\n${fragments.map((frag) => print(frag)).join('\n')}`;
```
JoviDeCroock, line 137: This might not 100% match the hash, not sure if that's a problem. Can try to align it more with the LSP

JoviDeCroock:
```suggestion
        const document = `${operation}\n\n${fragments.map((frag) => print(frag)).join('\n\n')}`;
```

packages/cli-utils/src/index.ts
```diff
@@ -132,6 +133,10 @@ async function main() {
         output: options.output,
       });
     })
+    .command('generate-persisted <target>')
+    .action(async (target) => {
+      await generatePersisted(target || './po.json');
```
JoviDeCroock, line 138:
```suggestion
      await generatePersisted(target);
```

PR #183, opened 2024-04-10
feat(cli): cache typescript types

### LINKED ISSUE #160: RFC: Implement codegen command to use AOT-compiled, cached types (Turbo Mode)
## Context

Currently, `gql.tada` is an alternative to codegen tools, which generate types, like the [`client-preset`](https://the-guild.dev/graphql/codegen/plugins/presets/preset-client).

Instead of assigning pre-generated types to GraphQL documents, they're instead parsed and inferred in TypeScript’s type system itself. However, this increases `tsc` type checking time at least linearly with the amount of documents a user writes and creates.
While this also is a problem with codegen tools (the time of tools run in parallel in a watch mode, or ahead of time, will increase as well), the processing that type inference takes will almost always be much larger.

In #150, we've experimented with [printing internal types from TypeScript in the CLI](https://github.com/0no-co/gql.tada/blob/48aeae24f67cfdace6badb2e55dd30eb52b06690/packages/internal/src/introspection.ts#L62-L83).

## Summary

Instead of always inferring the type of GraphQL documents, we could instead allow `graphql()` to lookup pre-computed types, which are created ahead of time.

This would mean that `graphql()` has two ways of looking up a type for a given document:
- It first checks whether a cache contains the query its been passed and uses that type if it's available
- If not; it falls back to inferring the type of the given document on the fly

The CLI would be used to populate the cache that `graphql()` reads from (#76). This is dependent on the CLI having an implementation to discover documents, which is necessary for #77 anyway.

**tl;dr: This means that a user can, after they're done editing queries and getting type updates in real-time, run the CLI and commit ahead-of-time-precompiled types for all queries.**

Since this is a transparent optimisation, this is like a **turbo** mode for `gql.tada` and merges the approaches of codegen and inferred GraphQL types and creates a seamless bridge.

This should also alleviate performance problems when compared to other codegen approaches indefinitely.

## Proposed Solution

- `gql.tada` has a global, declaration-based queries cache with types
- `graphql()` can lookup types in this cache and use them
- The CLI has a command (name TBD) to generate this cache

JoviDeCroock:
Resolves #160 

## Summary

This compiles types Ahead Of Time so that we can have a cache and that during dev TypeScript has less to worry about, current output looks like:

```
{
  'graphql(`\r\n  fragment PokemonItem on Pokemon {\r\n    id\r\n    name\r\n  }\r\n`)': 'TadaDocumentNode<{ name: string; id: string; }, {}, { fragment: "PokemonItem"; on: "Pokemon"; masked: true; }>',
  'graphql(`\r\n  query Pokemons ($limit: Int = 10) {\r\n    pokemons(limit: $limit) {\r\n      id\r\n      ...PokemonItem\r\n    }\r\n  }\r\n`, [PokemonItemFragment])': 'TadaDocumentNode<{ pokemons: ({ [$tada.fragmentRefs]: { PokemonItem: "Pokemon"; }; id: string; } | null)[] | null; 
}, { limit?: number | null | undefined; }, void>'
}
```

### DIFF
--- a/packages/cli-utils/src/commands/cache.ts
+++ b/packages/cli-utils/src/commands/cache.ts
@@ -1,0 +1,71 @@
+import { Project, ts } from 'ts-morph';
+import path from 'node:path';
+
+import { getTsConfig } from '../tsconfig';
+import type { GraphQLSPConfig } from '../lsp';
+import { getGraphQLSPConfig } from '../lsp';
+import { createPluginInfo } from '../ts/project';
+
+export async function generateGraphQLCache() {
+  const tsConfig = await getTsConfig();
+  if (!tsConfig) {
+    return;
+  }
+
+  const config = getGraphQLSPConfig(tsConfig);
+  if (!config) {
+    return;
+  }
+
+  await getPersistedOperationsFromFiles(config);
+}
+
+async function getPersistedOperationsFromFiles(
+  config: GraphQLSPConfig
+): Promise<Record<string, string>> {
+  // TODO: leverage ts-morph tsconfig resolver
+  const projectName = path.resolve(process.cwd(), 'tsconfig.json');
+  const project = new Project({
+    tsConfigFilePath: projectName,
+  });
+
+  const pluginCreateInfo = createPluginInfo(project, config, projectName);
+
+  const sourceFiles = project.getSourceFiles();
+
+  return sourceFiles.reduce((acc, sourceFile) => {
+    const tadaCallExpressions = findAllCallExpressions(sourceFile.compilerNode, pluginCreateInfo);
+    return {
+      ...acc,
+      ...tadaCallExpressions.reduce((acc, callExpression) => {
+        const typeChecker = project.getTypeChecker().compilerObject;
+        const resolvedSignature = typeChecker.getResolvedSignature(callExpression);
+        if (!resolvedSignature) {
+          return acc;
+        }
+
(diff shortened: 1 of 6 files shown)


packages/cli-utils/src/commands/cache.ts
```diff
@@ -0,0 +1,106 @@
...
+        if (!resolvedSignature) return acc;
+
+        const returnType = resolvedSignature.getReturnType();
+        const typeNode = typeChecker.typeToTypeNode(returnType, undefined, BUILDER_FLAGS);
+        if (!typeNode) return acc;
+        acc[callExpression.arguments[0].getText().slice(1, -1)] = printer.printNode(
+          EmitHint.Unspecified,
+          typeNode,
+          sourceFile.compilerNode
+        );
```
kitten, line 76: I found that using a manual `printer` + `printer.printNode` can lead to some omitted types and strings, especially if `typeNode` gets resinstantiated. There are some ways to accidentally let the checker reinstantiate types, which can lead to only partially evaluated types.

Might then also be missing `BUILDER_FLAGS` on top

packages/cli-utils/src/index.ts
```diff
@@ -99,6 +100,10 @@ prog.version(process.env.npm_package_version || '0.0.0');
 
 async function main() {
   prog
+    .command('cache')
```
JoviDeCroock, line 103:
```suggestion
    .command('turbo')
```

PR #184, opened 2024-04-10
feat: Allow enums to be remapped like scalars

### LINKED ISSUE #180: RFC: Allow enums to be remapped like scalars
## Summary

It would be great if there was a configuration option to make `gql.tada` emit typescript enums instead of union types for graphql enums.

There is a huge difference between these, one of them is that a typescript enum is an object that gets it into the transpiled js code, unlike a union type.

Moreover, this would imrpove compatibility of `tada.gql` with other existing codegen libs, namely `@graphql-codegen`. This, in turn, would ease the adoption and migration process for existing projects. In my particular case, the lack of this feature is an actual blocker for migrating my existing projects to `gql.tada`.

## Proposed Solution

Let’s say we have a following piece of graphql schema:

```
query {
  applicant {
    status: APPLICANT_STATUS
  }
}
```

Schema interpolation should return something like this:

```
export enum APPLICANT_STATUS {
  ONLINE: "ONLINE",
  OFFLINE: "OFFLINE",
  BLOCKED: "BLOCKED"
}

export type Applicant {
  status: APPLICANT_STATUS;
}
```

Whereas currently it returns

```
export enum APPLICANT_STATUS {
  ONLINE: "ONLINE",
  OFFLINE: "OFFLINE",
  BLOCKED: "BLOCKED"
}

export type Applicant {
  status: "ONLINE" | "OFFLINE" | "BLOCKED";
}
```

kitten:
Resolves #180

## Summary

This allows enums to be remapped like scalars. Specifically, if a user has a more qualified type for enums, they should be allowed to completely remap it.

Because `Omit<Schema['types'], keyof Scalars>` is potentially expensive for TypeScript to traverse, this PR avoids this pattern and instead refactors the code that resolves the types for enums. Specifically, `mapEnum` now outputs its type onto the `enumValues` property rather than `type`.

This PR also removes some redundant checks throughout the touched code path.

In practice, the following is now possible:

```ts
enum TestEnum {
  value = 'VALUE',
  test = 'TEST',
}

const graphql = initGraphQLTada<{
  introspection: someIntrospection;
  scalars: {
    // enum Test { VALUE, TEST }
    Test: TestEnum;
  };
}>();
```

In the future, we could add an additional check to `graphql.scalar` that checks the remapped type against the `enumValues` type to make sure they're compatible.

## Set of changes

- Refactor `getInputObjectTypeRec` and `_getVariablesRec` in `variables.ts` to move `obj<>` squashing
- Remove `SCALAR` and `ENUM` markers from processed schema outputs
- Remove redundant `unknown | never` output from processed schema outputs
- Slim down introspection generics once again
- Allow `scalars` to be an object of `readonly` properties
- **Output enum type onto `enumValues` and modify `variables.ts` and `selection.ts` to check `type` then `enumValues`**

### DIFF
--- a/.changeset/fair-otters-lie.md
+++ b/.changeset/fair-otters-lie.md
@@ -1,0 +1,6 @@
+---
+"gql.tada": minor
+"@gql.tada/internal": patch
+---
+
+Allow GraphQL enum types to be remapped with the `scalars` configuration option.
--- a/packages/internal/src/introspection/preprocess.ts
+++ b/packages/internal/src/introspection/preprocess.ts
@@ -57,7 +57,7 @@ const printFields = (fields: readonly IntrospectionField[]) => {
 export const printIntrospectionType = (type: IntrospectionType) => {
   if (type.kind === 'ENUM') {
     const values = printNamedTypes(type.enumValues);
-    return `{ name: ${printName(type.name)}; type: ${values}; }`;
+    return `{ name: ${printName(type.name)}; enumValues: ${values}; }`;
   } else if (type.kind === 'INPUT_OBJECT') {
     const fields = printInputFields(type.inputFields);
     return `{ kind: 'INPUT_OBJECT'; name: ${printName(type.name)}; inputFields: ${fields}; }`;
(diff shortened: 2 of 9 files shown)


VinceBT-BG:
Can this be done automatically for any enums that is present in the schema ? Instead of having to list them one by one

kitten:
@VinceBT-BG: You can write a script for this. But mainly, please check the related discussions for more info ✌️

PR #188, opened 2024-04-11
feat: Add support for the second argument to `graphql.persisted`

JoviDeCroock:
## Summary

This adds support for `graphql.persisted(hash, document)` so we don't have to advise folks to use APQ when their GraphQL client does not support definition-less documents. To push this out we'll still need to add support to the LSP.

### DIFF
--- a/.changeset/late-mice-type.md
+++ b/.changeset/late-mice-type.md
@@ -1,0 +1,6 @@
+---
+"@gql.tada/cli-utils": minor
+"gql.tada": minor
+---
+
+Support a second argument in `graphql.persisted` which represents the `DocumentNode` to create the possibility of persisted-operations that still lock down the origin. Before this change we had to advise folks to use APQ.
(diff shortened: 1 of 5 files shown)


kitten (approved):

.changeset/late-mice-type.md
```diff
@@ -0,0 +1,6 @@
+---
+"@gql.tada/cli-utils": minor
+"gql.tada": minor
+---
+
+Support a second argument in `graphql.persisted` which represents the `DocumentNode` to create the possibility of persisted-operations that still lock down the origin. Before this change we had to advise folks to use APQ.
```
kitten, line 6:
```suggestion
Support a second argument in `graphql.persisted` which accepts a `TadaDocumentNode` rather than passing a generic. This allows the document node to not be hidden, to still generate `documentId` via `gql.tada` without having to hide the document during runtime.
```

Issue #189, opened 2024-04-11
Support query fails to cover `includeDeprecated` arg


hpohlmeyer:
### Describe the bug

When generating a schema via the CLI, I get the error:

> Something went wrong while trying to load the schema. CombinedError: [GraphQL] Field 'args' doesn't accept argument 'includeDeprecated'

I did a little bug-hunting and was on a similar route to #142. The support query is supposed to ensure that the introspection query does not use properties that are not supported by the schema. Unfortunately the support query does not cover this case.

---

**Bug cause**

I have recreated the `IntrospectSupportQuery` and tried it against my endpoint (provided by DatoCMS).

<details>
<summary>IntrospectSupportQuery</summary>

```gql
query IntrospectSupportQuery {
  directive: __type(name: "__Field") {
    fields {
      name
    }
  }
    type: __type(name: "__Type") {
    fields {
      name
    }
  }
  inputValue: __type(name: "__InputValue") {
    fields {
      name
    }
  }
}
```

</details>

<details>
<summary>IntrospectionSupportQuery Results</summary>

```json
{
  "data": {
    "directive": {
      "fields": [
        {
          "name": "args"
        },
        {
          "name": "deprecationReason"
        },
        {
          "name": "description"
        },
        {
          "name": "isDeprecated"
        },
        {
          "name": "name"
        },
        {
          "name": "type"
        }
      ]
    },
    "type": {
      "fields": [
        {
          "name": "description"
        },
        {
          "name": "enumValues"
        },
        {
          "name": "fields"
        },
        {
          "name": "inputFields"
        },
        {
          "name": "interfaces"
        },
        {
          "name": "kind"
        },
        {
          "name": "name"
        },
        {
          "name": "ofType"
        },
        {
          "name": "possibleTypes"
        }
      ]
    },
    "inputValue": {
      "fields": [
        {
          "name": "defaultValue"
        },
        {
          "name": "deprecationReason"
        },
        {
          "name": "description"
        },
        {
          "name": "isDeprecated"
        },
        {
          "name": "name"
        },
        {
          "name": "type"
        }
      ]
    }
  }
}
```

</details>

As you can see it includes `inputValue.fields.isDeprecated`, [which means `inputValueDeprecation` is marked as supported](https://github.com/0no-co/gql.tada/blob/main/packages/internal/src/loaders/query.ts#L34).

[...]

PR #191, opened 2024-04-12
fix: add support for detecting argument deprecation

JoviDeCroock:
Resolves https://github.com/0no-co/gql.tada/issues/189

## Summary

It looks like `InputValueDeprecation` does not necessarily mean argument deprecation, the safe way to find this would be to look at the locations the directive can be specified on.

Hence our query now adds in

```
schema: __schema { directives { name locations } }
```

EDIT: I realised that both args and input-values are of type `__InputValue` which would be a really odd inconsistency in datocms

### DIFF
--- a/.changeset/friendly-beans-exist.md
+++ b/.changeset/friendly-beans-exist.md
@@ -1,0 +1,5 @@
+---
+"@gql.tada/internal": patch
+---
+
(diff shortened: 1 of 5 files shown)


JoviDeCroock:
Issue in question just has a non-spec compliant server I think

PR #201, opened 2024-04-15
chore(docs): general CI/Development workflow guide

JoviDeCroock:
## Summary

This adds a few workflows for CI/pre-commit: for checking whether we are still up-to-date, diagnostics in CI and caching the types.

I chose not to go into `generate-persisted` yet and maybe even put that into its own document, not sure yet?


website/guides/workflow.md
```diff
@@ -0,0 +1,94 @@
...
+---
+
+# Workflow
+
+There are some general workflows that are distinct locally (i.e. the LSP can take care of it)
+vs remotely. In this guide we'll look at how we can get local/remote to being consistent.
+
+> [!NOTE]
+>
+> This assumes that every `gql.tada` file is checked into VC.
```
kitten, line 13: Hm, I'm not sure what the note is getting at, it doesn't seem to prompt or warn the user of anything, and seems a little hard to read. (VC is also not a common abbreviation over "checked into the repository")

Should we remove this and make case-by-case evaluations of what the trade-offs are of checking files in or not later on?

website/guides/workflow.md
```diff
@@ -0,0 +1,94 @@
+---
+title: Workflow
+description: How does gql.tada and its LSP fit into your wofklow?
+---
+
+# Workflow
+
+There are some general workflows that are distinct locally (i.e. the LSP can take care of it)
+vs remotely. In this guide we'll look at how we can get local/remote to being consistent.
```
kitten, line 9: > locally (i.e. the LSP can take care of it) [...] vs remotely

I don't think that's a complete picture here and might not be relevant immediately when clicking into this docs page.

Re. the terminology in general; I think “locally” includes the CLI, so I'd differentiate between:
- installation (excluded)
- opening a file and editing it (could be represented by a section linking to the GraphQLSP reference API doc)
- persisted (excluded)
- `generate output` if `tsc` is run but the editor is not
- saving the file and updating turbo
- remote: CI

website/guides/workflow.md
```diff
@@ -0,0 +1,94 @@
...
+An example could look like the following
+
+```yaml
+- name: Verify schema and tada types are up to date
+  run: |
+    npx gql-tada generate-schema http://example.com
+    npx gql-tada generate-output -c ./tsconfig.json
+    npx gql-tada generate-persisted ./persisted-ops.json
+    git diff --name-only --exit-code -- .
+```
```
kitten, line 33: `generate-persisted` is unrelated here, so I'd remove it? It's more something that's needed to be passed to other scripts/tasks

kitten: There's no indication here that this is using GitHub workflow instance.

I'd maybe pull all workflow examples into a sub-section at the end of this page, and just keep the commands generic here.

kitten: Re. `generate-schema`, also potentially not relevant here, since it only applies when the schema isn't easily accessible, so let's maybe keep that out of a CI-leaning section?

I think we can have a section above re. external/remote schemas (i.e. schemas on URLs that aren't easily accessible or are in separate repos with the recommendation to modify the config to a local SDL file and to use the CLI)

kitten (commented):
I think this generally showcases GitHub Workflows, but is a little structured towards local usage, which is a bit of a conflict right now 🤔 Didn't have enough of a look to have a broader idea though on what to change yet

PR #225, opened 2024-04-17
fix: gql.tada generate --output *.ts was outputting .d.ts

llllvvuu:
If you use `gql.tada generate --output introspection.ts` and there is no tadaOutputLocation in tsconfig.json, .d.ts contents will be output into a .ts file. Then initGraphQLTada will fail.

### DIFF
--- a/.changeset/silent-apes-shop.md
+++ b/.changeset/silent-apes-shop.md
@@ -1,5 +1,5 @@
 ---
-"@gql.tada/internal": patch
+"@gql.tada/cli-utils": patch
 ---
 
-Fix CLI outputting .d.ts when .ts input file specified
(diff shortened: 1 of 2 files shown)


kitten (approved):
nice catch!

.changeset/silent-apes-shop.md
```diff
@@ -0,0 +1,5 @@
+---
+"@gql.tada/internal": patch
+---
+
+Fix CLI outputting .d.ts when .ts input file specified
```
kitten, line 5:
```suggestion
"@gql.tada/cli-utils": patch
---

Fix `generate-output` command outputting the `.d.ts` format when `.ts` extension was specified instead
```

PR #232, opened 2024-04-19
feat(cli-utils): add experimental support for `.vue` SFC files

### LINKED ISSUE #210: `gql.tada check` reports an "Unkown fragment" error when fragments are imported in a TS file from a Vue SFC
### Describe the bug

We're trying to setup gql.tada to do a check in a CI job using the newly added check command (awesome addition btw!) but we're faced with a weird issue. The command will report incorrectly an error when fragments are imported from a different file type. This is what I've tested:

- ✅ Fragment imported in a TS file from another TS file
- ✅ Fragment imported in a Vue component from another Vue component file
- ❌ Fragment imported in a TS file from a Vue component file

The IDE however, is fine with any kind of combination. I'm unsure where the problem lies exactly but it seems strange that the IDE would not report the error if both the CLI and IDE are using the LSP?

I've set up a reproduction that hopefully will help you track the source down.

Thank you so much for the awesome project, it might be the best package I've used in a long time!

### Reproduction

https://github.com/raphael-yapla/gql-tada-vue-issue

### gql.tada version

gql.tada v1.5.0

### Validations

### LINKED ISSUE #231: gql.tada turbo doesn't generate cache for documents in Vue SFC files
### Describe the bug

It seems like gql.tada turbo does not work with Vue SFC files. Cache entries in my project are only generated for normal typescript files, but none for any documents defined in my components.

See the reproduction provided for a minimal example. The graphql-cache.d.ts file only contains the query of `random.ts`, although there is another query defined in `PokemonList.vue`.

Maybe i misunderstood something, but i assume that those should be generated as well?

Thank you!

### Reproduction

https://github.com/KammererTob/gql.tada-turbo-bug/tree/main

### gql.tada version

gql.tada v1.5.3

### Validations

JoviDeCroock:
Resolves #210 
Resolves #231

## Summary

This PR adds support for parsing `.vue` files, it will use the Vue compiler to create an intermediary representation in `ts-morph`.

Screenshot 2024-04-19 at 10 15 39

I added this initially only to `check` so we can align on the solution and I can expand it into multiple other after.

Some potential improvements

- use the `moduleResolutionHost` to route `.vue` files to `.vue.tada.ts` or something similar. Now we are forced to name them `.vue.ts` files so that TS picks them up automatically


packages/cli-utils/src/ts/project.ts
```diff
@@ -1,5 +1,57 @@
...
+      return rm(sourceFile.compilerNode.fileName);
+    })
+  );
+};
+
+export const polyfillVueSupport = async (
+  project: Project,
+  ts: typeof import('typescript/lib/tsserverlibrary')
+): Promise<Array<SourceFile>> => {
+  const vueProjectFiles = project.addSourceFilesAtPaths('./src/**/*.vue');
```
JoviDeCroock, line 18: This should be replaced by checking the `include` folders

packages/cli-utils/src/ts/project.ts
```diff
@@ -1,5 +1,59 @@
 import type { GraphQLSPConfig } from '@gql.tada/internal';
-import type { Project } from 'ts-morph';
+import type { Project, SourceFile } from 'ts-morph';
+import * as vue from '@vue/language-core';
```
JoviDeCroock, line 3: Could be an optional peer-dep I reckon?

examples/example-pokemon-vue/po.json
```diff
@@ -0,0 +1,3 @@
+{
+  "sha256:fc073da8e9719deb51cdb258d7c35865708852c5ce9031a257588370d3cd42f3": "\n      query Pokemons($limit: Int = 10) {\n        pokemons(limit: $limit) {\n          id\n          ...PokemonItem\n        }\n      }\n    "
```
JoviDeCroock, line 2: This is still wrong as it hasn't defined the `PokemonItem` fragment

examples/example-pokemon-vue/src/graphql-cache.d.ts
```diff
@@ -0,0 +1,12 @@
+/* eslint-disable */
+/* prettier-ignore */
+import type { TadaDocumentNode, $tada } from 'gql.tada';
+
+declare module 'gql.tada' {
+ interface setupCache {
+    "\n      query Pokemons($limit: Int = 10) {\n        pokemons(limit: $limit) {\n          id\n          ...PokemonItem\n        }\n      }\n    ":
+      TadaDocumentNode<{ pokemons: ({ [x: string]: any; } | null)[] | null; }, { limit?: number | null | undefined; }, void>;
```
JoviDeCroock, line 8: This is wrong as it's typing the pokemon as any

packages/cli-utils/src/commands/turbo/thread.ts
```diff
@@ -27,18 +28,26 @@ async function* _runTurbo(params: TurboParams): AsyncIterableIterator<TurboSigna
...
-    scriptKind: ScriptKind.TS,
-  });
+  const overrideFile = project.createSourceFile(
+    '__gql-tada-override__.d.ts',
+    DECLARATION_OVERRIDE,
+    {
+      overwrite: true,
+      scriptKind: ScriptKind.TS,
+    }
+  );
```
kitten, line 38: side-note: seems very odd that this has changed, since the prior wasn't reformatted by Prettier before

packages/cli-utils/src/commands/turbo/thread.ts
```diff
@@ -85,6 +94,12 @@ async function* _runTurbo(params: TurboParams): AsyncIterableIterator<TurboSigna
       warnings,
     };
   }
+
+  await rm(overrideFile.compilerNode.fileName);
```
kitten, line 98: If we're actually creating these files, in theory, just as `__gql-tada-override__.d.ts` above, we should aim to create these virtually 🤔

It'd likely be good to abstract this part anyway into a small helper.

Generally the approach of:
- create `ts-morph` project with `skipAddingFilesFromTsConfig: true`
- adding virtual files
- `addSourceFilesFromTsConfig`
Should be abstractable in theory and allow us to reuse this across all commands without forgetting it

examples/example-pokemon-vue/po.json
```diff
@@ -0,0 +1,3 @@
+{
```
kitten, line 1: Let's make sure not to remove these files before merging this, like in the other example.

We could add their names to the config if we really wanted them, but I think we won't need to commit them

packages/cli-utils/src/ts/project.ts
kitten: This depends. The `doctor` command can in theory prompt people who use Vue to install `@vue/language-core` as a direct dependency in their project. Depends on whether that's typically already done by people commonly

packages/cli-utils/src/ts/project.ts
JoviDeCroock:
```suggestion
  const vueProjectFiles = project.addSourceFilesAtPaths('./**/*.vue');
```

packages/cli-utils/src/ts/project.ts
```diff
@@ -1,5 +1,48 @@
...
+
+export const polyfillVueSupport = async (
+  project: Project,
+  ts: typeof import('typescript/lib/tsserverlibrary')
+): Promise<Array<SourceFile>> => {
+  // TODO: scope this to the include array
```
JoviDeCroock, line 9:
```suggestion
```

kitten (approved):

PR #236, opened 2024-04-20
fix(cli-utils): Use source maps to correct offsets for Vue SFC files

### LINKED ISSUE #235: `gql.tada check` reports an "Unknown fragment" error when using nested fragment composition in Vue SFCs
### Describe the bug

This is a follow up to #210.

First of, a big thank you for adding Vue support so quickly! After setting it up on our bigger project I found a blindspot in my previous reproduction repo. Nested fragment composition doesn't seem to be working with Vue right now. Again Volar and the IDE seem very happy with everything but the command `gql.tada check` gives a "Unknown fragment" error again.

I've made a new branch on the reproduction repository highlighting the new error. Please let me know if you need any more details!

### Reproduction

https://github.com/raphael-yapla/gql-tada-vue-issue/tree/nested-fragment-composition

### gql.tada version

gql.tada 1.5.4

### Validations

JoviDeCroock:
Resolves #235 

## Summary

It looks like with a minimal project the offsets are contained enough to just work, however the more files we add the larger this offsets seems to become. This PR introduces support for looking up positions in the `vue.Sourcemap` so we can triangulate our position in the accompanying `TypeScript` file.


.changeset/lucky-brooms-draw.md
```diff
@@ -0,0 +1,5 @@
+---
+"@gql.tada/cli-utils": patch
+---
+
+Leverage the Sourcemap the Vue compiler delivers us when looking up positions in the TS document
```
kitten, line 5:
```suggestion
Use the Vue SFC compiler's source-maps when looking up positions in its TypeScript file output
```

examples/example-pokemon-vue/src/graphql-env.d.ts
```diff
@@ -15,17 +15,17 @@ export type introspection = {
   subscription: never;
   types: {
     'Attack': { kind: 'OBJECT'; name: 'Attack'; fields: { 'damage': { name: 'damage'; type: { kind: 'SCALAR'; name: 'Int'; ofType: null; } }; 'name': { name: 'name'; type: { kind: 'SCALAR'; name: 'String'; ofType: null; } }; 'type': { name: 'type'; type: { kind: 'ENUM'; name: 'PokemonType'; ofType: null; } }; }; };
+    'Int': unknown;
```
kitten, line 18: nit: Would be good to check why this changed unless it's just copied over or to skip these changes

kitten (approved):

PR #241, opened 2024-04-21
feat(cli-utils): add experimental support for `.svelte` files

kitten:
Related: https://github.com/0no-co/gql.tada/pull/232
Related: https://github.com/0no-co/gql.tada/pull/236

## Summary

This abstracts our usage of `volar`’s utilities down to `VirtualCode` wrappers and abstracts it to have parity with the Svelte file support.
The Svelte2TSX compiler has been added to compile Svelte files to TSX as well.

Notably, this changes the approach slightly. To risk confusing external files with virtual files, we now create a `SourceFile` of the project file (svelte/vue file) outside of our TypeScript program and don't add it to the project.

Redirections to the virtual file and such still function as before. But the biggest change is that all commands will now iterate over the virtual files (e.g. `.vue.ts` and `.vue.svelte`) instead of the non-TS project files (`.vue` and `.svelte`), since those aren't loaded into TypeScript anymore. The output of the logger is hence mapped and `getFilePosition` now uses the position mapper as needed.

## Set of changes

- Squash usage of Volar helpers down to just `VirtualCode`
- Add async-imports for external virtual code creators
- Add file location mapper
  - This will either map from `project -> virtual` or `virtual -> project` file, so it can be used for both, the GraphQLSP plugin info methods, and the logger outputs
- Add svelte example


packages/cli-utils/src/ts/virtualCode/index.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    if (virtualSourceFile._markAsInProject) virtualSourceFile._markAsInProject();
+
+    projectToVirtual.set(file.fileId, virtualSourceFile.compilerNode);
+    virtualToProject.set(virtualFileId, projectSourceFile);
+    sourceMaps.set(file.fileId, sourceMap);
+    sourceMaps.set(virtualFileId, sourceMap);
+  }
+
+  return ((fileId, position) => {
```
kitten, line 81: We're still unable to override the original filename with the output here, which would allow us to clean that up. Since TypeScript does support arbitrary file extensions, I'm not quite sure where that issue may be coming from, but for now, this workaround is still needed

packages/cli-utils/src/ts/virtualCode/vue.ts
```diff
@@ -0,0 +1,35 @@
+import type { VirtualCode } from '@vue/language-core';
+import { forEachEmbeddedCode, getDefaultVueLanguagePlugins } from '@vue/language-core';
+import * as vueCompilerDOM from '@vue/compiler-dom';
+import * as vue from '@vue/language-core';
+
+import type { CreateVirtualCode } from './types';
+
+const VueVirtualCode =
+  vue.VueGeneratedCode || ((vue as any).VirtualCode as typeof vue.VueGeneratedCode);
```
kitten, line 9: This export was renamed on the `main` branch of `@vuejs/language-tools`, so both names will have to be checked here to anticipate their next release: https://github.com/vuejs/language-tools/commit/f46634c7e5a330032b40d224cf72dd1c0b6640d5

JoviDeCroock (approved):

PR #247, opened 2024-04-23
refactor(cli-utils): Refactor internal TypeScript instantiation

kitten:
## Summary

This refactors the internal instantiation of `typescript` from using `ts-morph` to `@typescript/vfs`, as a lighter, more reviewable alternative that will suffice for our purposes for the time being and for foreseeable use-cases.

We're essentially aiming to replace this to:
- increase correctness of instantiating a TypeScript environment
- improve virtua file loading and mapping
- potentially gain more control for when we need to replace internal TypeScript mechanisms (e.g. custom resolution)
- increase accuracy of "fake" `PluginCreateInfo` environment

## Set of changes

- Replace `PluginCreateInfo` environment
  - Polyfill virtual file transformer support into more `LanguageService` methods
  - Increase correctness of `Program` polyfills
- Refactor `.vue` and `.svelte` transformers to be simpler
- Add `programFactory` to assemble a program
- Retrieve lib files from `node_modules`
- Consistently default-import from `typescript`
  - ESM support is otherwise broken


packages/cli-utils/src/commands/shared/logger.ts
```diff
@@ -54,6 +54,9 @@ export function externalError(message: string, error: unknown) {
     ) {
       title = 'code' in error ? 'System Error' : 'Error';
       text = (error as Error).message.trim();
+    } else if ('stack' in error && typeof error.stack === 'string') {
```
kitten, line 57: Quick note: I'm just committing this, we always need it in development, and there's no harm in seeing more output when someone reports an issue

JoviDeCroock (approved):

PR #251, opened 2024-04-24
fix(cli-utils, internal): Fix retrieval of `tsconfig.json` extends and standard TS libs

kitten:
- Fix resolution of `extends` options
- Fix incorrect `bind` on `getSourceFile`
- Fix resolution of lib files from non-standard paths

### DIFF
--- a/packages/cli-utils/src/ts/factory.ts
+++ b/packages/cli-utils/src/ts/factory.ts
@@ -49,19 +49,17 @@ export const programFactory = (params: ProgramFactoryParams): ProgramFactory => 
   const virtualMap: VirtualMap = new Map();
 
   const projectRoot = path.dirname(params.configPath);
-  const tslibPath = path.join(projectRoot, 'node_modules/typescript/lib/');
-  const system = createFSBackedSystem(vfsMap, projectRoot, ts, tslibPath);
+  const system = createFSBackedSystem(vfsMap, projectRoot, ts);
   const config = resolveConfig(params, system);
-
   for (const { filename, contents } of resolveLibs(params)) {
-    if (contents) system.writeFile(path.join(tslibPath, filename), contents);
+    if (contents) system.writeFile('/' + filename, contents);
   }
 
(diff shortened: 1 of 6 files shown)


packages/cli-utils/src/ts/factory.ts
```diff
@@ -49,19 +49,17 @@ export const programFactory = (params: ProgramFactoryParams): ProgramFactory =>
...
   const config = resolveConfig(params, system);
-
   for (const { filename, contents } of resolveLibs(params)) {
-    if (contents) system.writeFile(path.join(tslibPath, filename), contents);
+    if (contents) system.writeFile('/' + filename, contents);
   }
 
   const rootNames = new Set(config.fileNames);
   const options = {
+    getDefaultLibFilePath: ts.getDefaultLibFilePath(config.options),
```
kitten, line 60: Note: this is basically a last-resort location, but TS often doesn't use this 🤷

packages/cli-utils/src/ts/factory.ts
```diff
@@ -49,19 +49,17 @@ export const programFactory = (params: ProgramFactoryParams): ProgramFactory =>
...
 
   const projectRoot = path.dirname(params.configPath);
-  const tslibPath = path.join(projectRoot, 'node_modules/typescript/lib/');
-  const system = createFSBackedSystem(vfsMap, projectRoot, ts, tslibPath);
+  const system = createFSBackedSystem(vfsMap, projectRoot, ts);
   const config = resolveConfig(params, system);
-
   for (const { filename, contents } of resolveLibs(params)) {
-    if (contents) system.writeFile(path.join(tslibPath, filename), contents);
+    if (contents) system.writeFile('/' + filename, contents);
```
kitten, line 55: `@typescript/vfs` seems to expect that the lib files are at the virtual root folder, even if an override is passed to `createFSBackedSystem`. The former's argument for this doesn't seem to be used correctly

JoviDeCroock (approved):

PR #276, opened 2024-04-28
feat(cli): add `gql.tada/ts-plugin` to init/doctor

JoviDeCroock:
Relates to #75

## Summary

This PR adds a new minimum version for TypeScript to support sub-modules in a library for the LSP Plugins. 

- In `doctor` we will check for `gql.tada/ts-plugin` when TS is > 5.5 so folks are pushed to migrate
- In `init` we automatically add `gql.tada/ts-plugin` when TS is > 5.5

### DIFF
--- a/packages/cli-utils/src/commands/doctor/runner.ts
+++ b/packages/cli-utils/src/commands/doctor/runner.ts
@@ -22,7 +22,7 @@ const delay = (ms = 700) => {
   }
 };
 
-const semiverComply = (version: string, compare: string) => {
+export const semiverComply = (version: string, compare: string) => {
   const match = version.match(/\d+\.\d+\.\d+/);
   return match ? semiver(match[0], compare) >= 0 : false;
 };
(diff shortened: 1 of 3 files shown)


packages/cli-utils/src/commands/doctor/runner.ts
```diff
@@ -94,21 +95,29 @@ export async function* run(): AsyncIterable<ComposeInput> {
...
-  } else if (!semiverComply(gqlspVersion[1], MINIMUM_VERSIONS.lsp)) {
-    yield logger.failedTask(Messages.CHECK_DEPENDENCIES);
-    throw logger.errorMessage(
-      `The version of ${logger.code('@0no-co/graphqlsp')} in your dependencies is out of date.\n` +
-        logger.hint(
-          `${logger.code('gql.tada')} requires at least ${logger.bold(MINIMUM_VERSIONS.lsp)}`
-        )
-    );
+  const supportsEmbeddedLsp = semiverComply(
+    typeScriptVersion[1],
```
kitten, line 99: Since TS is now a peer dependency, we should be able to read this from `require('typescript').version`. Probably have to try-catch it with a new error, but that should also just give us the raw version.

kitten (approved):

packages/cli-utils/src/commands/doctor/runner.ts
JoviDeCroock: Can do that in a follow up

PR #284, opened 2024-05-02
feat(tada): add isOneOf to the introspection type and support query

JoviDeCroock:
## Summary

Adds support for https://github.com/graphql/graphql-spec/pull/825

When we encounter an input marked as `isOneOf: true` we need to create a type that is an _or_ of all the properties spelled out in said input-object.

- In the support-query we check whether `__Type` contains a `isOneOf`
- In the `Introspection` type we add `isOneOf` to the properties
- Add new branch in `variables.ts` that differentiates between `|` and `&`

### DIFF
--- a/packages/internal/src/loaders/introspection.ts
+++ b/packages/internal/src/loaders/introspection.ts
@@ -16,6 +16,7 @@ export interface SupportedFeatures {
   inputValueDeprecation: boolean;
   directiveArgumentsIsDeprecated: boolean;
   fieldArgumentsIsDeprecated: boolean;
+  typesIsOneOf: boolean;
 }
 
 /** Data from a {@link makeIntrospectSupportQuery} result */
@@ -48,6 +49,7 @@ export const ALL_SUPPORTED_FEATURES: SupportedFeatures = {
   inputValueDeprecation: true,
   directiveArgumentsIsDeprecated: true,
   fieldArgumentsIsDeprecated: true,
+  typesIsOneOf: true,
 };
 
 export const NO_SUPPORTED_FEATURES: SupportedFeatures = {
@@ -56,12 +58,14 @@ export const NO_SUPPORTED_FEATURES: SupportedFeatures = {
   inputValueDeprecation: false,
   directiveArgumentsIsDeprecated: false,
   fieldArgumentsIsDeprecated: false,
+  typesIsOneOf: false,
 };
 
 /** Evaluates data from a {@link makeIntrospectSupportQuery} result to {@link SupportedFeatures} */
 export const toSupportedFeatures = (data: IntrospectSupportQueryData): SupportedFeatures => ({
   directiveIsRepeatable: _hasField(data.directive, 'isRepeatable'),
   specifiedByURL: _hasField(data.type, 'specifiedByURL'),
+  typesIsOneOf: _hasField(data.type, 'isOneOf'),
   inputValueDeprecation: _hasField(data.inputValue, 'isDeprecated'),
   directiveArgumentsIsDeprecated: _supportsDeprecatedArgumentsArg(data.directive),
   fieldArgumentsIsDeprecated: _supportsDeprecatedArgumentsArg(data.field),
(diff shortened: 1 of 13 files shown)


packages/internal/src/loaders/introspection.ts
```diff
@@ -56,12 +58,14 @@ export const NO_SUPPORTED_FEATURES: SupportedFeatures = {
...
   directiveArgumentsIsDeprecated: false,
   fieldArgumentsIsDeprecated: false,
+  typesIsOneOf: false,
 };
 
 /** Evaluates data from a {@link makeIntrospectSupportQuery} result to {@link SupportedFeatures} */
 export const toSupportedFeatures = (data: IntrospectSupportQueryData): SupportedFeatures => ({
   directiveIsRepeatable: _hasField(data.directive, 'isRepeatable'),
   specifiedByURL: _hasField(data.type, 'specifiedByURL'),
+  typesIsOneOf: _hasField(data.type, 'isOneOf'),
```
kitten, line 68: nit: I would name this `inputOneOf` as per the deprecation name for the other feature for example (drop the "is", change to "input" from type)

src/__tests__/variables.test-d.ts
```diff
@@ -36,6 +36,23 @@ describe('getVariablesType', () => {
...
+        toggleTodo (id: $id input: $input) { id }
+      }
+    `;
+
+    type doc = parseDocument<typeof query>;
+    type variables = getVariablesType<doc, schema>;
+
+    expectTypeOf<variables>().toEqualTypeOf<{
+      id: string;
+      // TODO: should we get rid of the leading empty object?
```
kitten, line 51: Yes, as per the RFC, exactly one property must be defined, i.e. all values are required and one must be specified when the input is passed

src/variables.ts
```diff
@@ -3,26 +3,36 @@ import type { SchemaLike } from './introspection';
...
+          Introspection,
+          true
+        > | null;
+      }
+  : {};
+
 type getInputObjectTypeRec<
   InputFields,
   Introspection extends SchemaLike,
+  IsOneOf extends boolean,
```
kitten, line 26: nit: I'd duplicate `getInputObjectTypeRec` here and hoist this check. I'm unsure whether infinitely mutually recursive inputs break if we don't

src/__tests__/variables.test-d.ts
```diff
@@ -36,6 +36,22 @@ describe('getVariablesType', () => {
...
+        toggleTodo (id: $id input: $input) { id }
+      }
+    `;
+
+    type doc = parseDocument<typeof query>;
+    type variables = getVariablesType<doc, schema>;
+
+    expectTypeOf<variables>().toEqualTypeOf<{
+      id: string;
+      input: { value_1: string } | { value_2: string };
```
kitten, line 51: Just discovered a bit of a wrench in the current implementation:
> Within the input object literal or unordered map, if the single entry is {null} an error must be thrown.

In the example that this PR provides, all input fields are defined as non-nullable, however, in the example that the RFC provides all fields are nullable. Meaning, it's irrelevant whether a field is non-nullable or nullable.

It seems like all input fields must be non-null.

src/variables.ts
```diff
@@ -3,6 +3,23 @@ import type { SchemaLike } from './introspection';
 import type { DocumentNodeLike } from './parser';
 import type { obj } from './utils';
 
+type GetInputObjectType<InputField, Introspection extends SchemaLike> = InputField extends {
+  name: any;
+  type: any;
+}
+  ? InputField extends { defaultValue?: undefined | null; type: { kind: 'NON_NULL' } }
+    ? {
+        [Name in InputField['name']]: unwrapTypeRec<InputField['type'], Introspection, true>;
```
kitten, line 12: As per the other comment, `true` here in `unwrapTypeRec` becoming `false` should fix the issue of non-null/null not mattering, as far as I remember.

Last review's comment suggested to remove `GetInputObjectType` entirely, to make sure we don't accidentally break mutually/infinitely recursive input types

PR #287, opened 2024-05-04
feat: Support globs and arrays of paths/globs for the `schema` option and `generate-schema` command arguments

kitten:
> [!NOTE]
> Marked as draft since it's kind of a proposal, due to it being requested multiple times. Coordination with `graphqlsp` is also needed.
> We may also want to use this opportunity to work on error handling (?)
> The risk of errors increases by **a lot** when we accept multiple SDL files, that have potentially not been processed by other tools.

## Summary

This has been requested multiple times now, and I thought I'd take a stab at implementing it.

When applied, this PR allows the `schema` option to accept arrays of paths, arrays of globs, and globs. This means that multiple SDL files can now be merged into a single one with the `generate-schema` command and that the SDL loader will discover multiple files.

This has been implemented using `ts.sys.readDirectory`, which should already support this pretty comprehensively.

When multiple `.json` files are discovered, an error is thrown.

The risk of errors increases by **a lot** when we accept multiple SDL files, that have potentially not been processed by other tools. This is because they might have been hand-written and not validated by any other tool. I haven't figured out yet what to do about this, especially since GraphQLSP/TSServer cannot output global errors

## Set of changes

- **DANGER:** The `loadFromSDL` config signature has changed to accept a `rootPath`
- `loadFromSDL` now uses `ts.sys.readDirectory` to discover multiple schema files
- When `loadFromSDL` is used with a single JSON file nothing changes, otherwise, if multiple files are discovered and one is a JSON file, an error is thrown
- `generate-schema` command now accepts multiple input strings

### DIFF
--- a/.changeset/angry-dancers-unite.md
+++ b/.changeset/angry-dancers-unite.md
@@ -1,0 +1,5 @@
+---
+"@gql.tada/cli-utils": minor
+---
+
+Allow `generate-schema` to accept globs or multiple paths/globs. The command will therefore now be able to merge separate SDL files into a single one.
--- a/.changeset/two-parents-tie.md
+++ b/.changeset/two-parents-tie.md
@@ -1,0 +1,5 @@
+---
+"@gql.tada/internal": minor
+---
+
+Support globs and arrays of paths/globs to be passed to the `schema` option. This allows the local schema to be extended or a schema with split SDL files to be configured with `gql.tada` without any extra tooling.
--- a/packages/cli-utils/src/commands/generate-schema/index.ts
+++ b/packages/cli-utils/src/commands/generate-schema/index.ts
@@ -24,9 +24,9 @@ const parseHeaders = (
 export class GenerateSchema extends Command {
   static paths = [['generate-schema'], ['generate', 'schema']];
 
-  input = Option.String({
+  input = Option.Rest({
     name: 'schema',
-    required: true,
+    required: 1,
   });
 
   tsconfig = Option.String('--tsconfig,-c', {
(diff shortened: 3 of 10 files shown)


DamianOsipiuk (commented):
Hi,

Will there also be support for an LSP plugin?

In my current use case, I’m relying on the introspection endpoint. That works, but it requires the server to be running before type definitions can be generated. Having a file-based approach with glob patterns would really simplify things for frontend authoring, since you wouldn’t need to ensure the server is up and running first.

Relying solely on introspection can also be tricky in some scenarios, for example:
- pipelines that validate whether generated `.d.ts` files were committed
- pre-push hooks for the same check

This would be super helpful. Really looking forward to it!

kitten:
Still unsure if it's worth adding this. Technically, all that'll do may be non-standard, and will work with a **small fraction** of GraphQL APIs. That's because this isn't the case:
> Having a file-based approach with glob patterns would really simplify things for frontend authoring, since you wouldn’t need to ensure the server is up and running first.
This is basically an assumption. Personally, I haven't used an SDL-based schema/API in years. For a long time, I've also used tagged-template-based GraphQL schemas, rather than SDL-file based ones. The nicest DX currently comes from code-first APIs/schemas (gqtx / pothos / nexus, et al)

That means, this is a very limited case. If we add this, the next question we'll get will certainly be around `#import`, transformations, and other special cases.

I was more interested in adding an example for SDL-on-start functionality, i.e. a small utility that writes the SDL output when the GraphQL server (as long as it's JS-based) starts (or re-starts). 

Writing the SDL file as the server runs is imho the nicest experience, since it's pretty trivial to commit the SDL file at that point too.

PR #291, opened 2024-05-10
passthrough query input string

wyattades:
## Summary

We'd like to send the query string directly to our backend, not the "parsed graphql definition" (not sure the terminology). We have a hacky patch workaround for this, but it seems like something appropriate for other users of this lib as well.

example:
```ts
const query = graphql(`
  query MyThing {
    id
  }
`);

const res = await requestDataFromBackend(query.input);
```

Question: i'm not currenctly using fragments, but it probably also makes sense to passhrough them as well?

## Set of changes

- passthrough the query input string

### DIFF
--- a/.changeset/yellow-singers-push.md
+++ b/.changeset/yellow-singers-push.md
@@ -1,0 +1,5 @@
+---
+'gql.tada': patch
+---
+
+passthrough the query input string
--- a/src/api.ts
+++ b/src/api.ts
@@ -233,7 +233,7 @@ function initGraphQLTada<const Setup extends AbstractSetupSchema>() {
       );
     }
 
-    return { kind: Kind.DOCUMENT, definitions };
+    return { kind: Kind.DOCUMENT, definitions, input };
   }
 
   graphql.scalar = function scalar(_schema: Schema, value: any) {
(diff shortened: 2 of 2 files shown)


kitten:
This is pretty non-standard and the canonical way of doing this is to simply `print` the query. Can you explain why you would need to do this? We really haven't come across this being necessary (and we've basically been working on `urql` for a while, and it's not really an issue)

wyattades:
> This is pretty non-standard and the canonical way of doing this is to simply `print` the query

Sorry I don't understand. Isn't the standard in the graphql world to send the query **string** to your backend?
Also, what does `print` the query mean?

>  it may make more sense for you to extend/wrap the graphql function, i.e. the initGraphQLTada returned function.

That's what we're doing now :) but I figure other people are also using gql.tada just for the type inference, and only accept graphql query strings on their backend.

kitten:
You use `print` from `graphql.js` (or `@0no-co/graphql.web` if you want a smaller alternative that `gql.tada`/`urql` are already using) and that gives you the GraphQL document in its printed, stringified form from the AST.

Most GraphQL clients will do this for you out of the box though, of course

wyattades:
> You use `print` from `graphql.js` (or `@0no-co/graphql.web` if you want a smaller alternative that `gql.tada`/`urql` are already using) and that gives you the GraphQL document in its printed, stringified form from the AST.
> 
> Most GraphQL clients will do this for you out of the box though, of course

I see your point, it would be more correct to stringify the AST. I'll close this PR

PR #322, opened 2024-06-17
docs: Add header hint for `generate-schema` to "Workflows" page

sistrall:
## Summary

When starting with gql.data, I spent quite some time trying to understand how to retrieve the schema without leaking API tokens in `tsconfig.json`.

## Set of changes

This is the solution I'm using in my projects. It proves to be quite effective and even simple, despite the convoluted usage of single and double quotes required to control when shell interpolates the variable.

It uses the `dotenv` command, which is pretty common to have in dev dependencies.

### DIFF
--- a/website/get-started/workflows.md
+++ b/website/get-started/workflows.md
@@ -41,6 +41,23 @@ use your configuration's `schema` setting, provided it's a file path.
   <p>Learn more about the <code>generate-schema</code> command</p>
 </a>
 
+`gql.tada generate schema` also comes useful when you
+need granular control on how to connect to a GraphQL endpoint. I.e. sometimes the API requires an authorization header with a token that you don't
+want to store in your reporitory. If that's the case, you can safely store the token in a `.env` file and use the `dotenv` to get the schema.
+
+::: code-group
+```json [package.json]
+{
+  "scripts": {
+    "generate-schema": "dotenv -c -- bash -c 'gql-tada generate-schema https://your.graphql.api.endpoint --header \"Authorization: $SOME_TOKEN_VIA_ENVIRONMENT_VARIABLE\"'",
+  },
+  "devDependencies": {
+    "dotenv-cli": "*"
+  }
+}
+```
+:::
+
 ---
 
 ### The `doctor` command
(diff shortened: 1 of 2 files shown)


kitten:
I don't think this is the right place for this per se. I think this could fit into a folded section of this doc's section however: https://gql-tada.0no.co/get-started/workflows#downloading-schemas

sistrall:
Thank you @kitten

I moved the info to the doc you suggested.

website/get-started/workflows.md
```diff
@@ -41,6 +41,23 @@ use your configuration's `schema` setting, provided it's a file path.
...
+{
+  "scripts": {
+    "generate-schema": "dotenv -c -- bash -c 'gql-tada generate-schema https://your.graphql.api.endpoint --header \"Authorization: $SOME_TOKEN_VIA_ENVIRONMENT_VARIABLE\"'",
+  },
+  "devDependencies": {
+    "dotenv-cli": "*"
+  }
+}
+```
+:::
```
kitten, line 59: (Side-note, this shouldn't be below the "CLI Reference" button as that links to the command as a closing "read more" link)

As noted here: https://github.com/0no-co/gql.tada/pull/322#issuecomment-2174283049
I have more concerns around these exact recommendations and the phrasing/language used here.

Basically, it's documentation is not an appropriate place to make exact recommendations like these when it's just trying to convey information.

In general, I'd assume that people who know how environment variables and `.env` files work will have no problems deriving this. If we present this exact `package.json` it just adds confusion. To put it generally, any code snippet that's in any library's documentation will eventually blindly be copied and this isn't important enough to just put it here as is. Sorry! ❤️ 

What I'd suggest is something that basically says:
- The `generate schema` CLI command also supports a `--header` option
- An API may require authentication to pull the schema
- A script can be written to specify an authentication header
- Just present a bare `gql.tada generate schema 'http://api.test/graphql' --header "Authorization: $YOUR_TOKEN"` command

The reasoning here is:
- `dotenv-cli`, `package.json:scripts` adds noise
- the bash wrapper is not cross-platform compatible and adds confusion for some users
- The URL/input to the `generate schema` command should match the prior example
- The `--output` I've omitted since it's already presented to decrease noise

Hinting that env variables + `--header` open up these options should be enough.

Sorry to be pedantic here! But it's a page in "Getting Started" so I want to make sure we get this right ❤️

Issue #323, opened 2024-06-17
gql.tada check erroneously reports field not in use in Vue SFC.


JarvisH:
### Describe the bug

Using the official Pokemon example for Vue:

```bash
 example-pokemon-vue git:(main) ✗ npx gql.tada check
⚠ Note: .vue and .svelte file support is experimental.

src/components/PokemonTypes.vue
  7:5	warn	Field types is not used.

src/components/PokemonItem.vue
  9:7	warn	Field name is not used.
```

The `name` field is reported as not being in use. Here is the template in question, were we can see `<p>{{ pokemon.name }}</p>` in use in the template. 

```vue
<script setup lang="ts">
import { FragmentOf, readFragment } from "../graphql";
import { ref, computed } from "vue";

const props = defineProps<{
  data: FragmentOf<typeof PokemonItemFragment> | null;
}>();

const pokemon = computed(() => readFragment(PokemonItemFragment, props.data));
</script>
<template>
  <li v-if="pokemon">
    <p>{{ pokemon.name }}</p>
    <PokemonTypes :data="pokemon" />
  </li>
</template>
```

Any fields referenced inside `<script setup>` are recognized however, only references inside `<template>` are ignored.

### gql.tada version

gql.tada 1.7.6

### Validations

- [X] I can confirm that this is a bug report, and not a feature request, RFC, question, or discussion, for which [GitHub Discussions](https://github.com/0no-co/gql.tada/discussions) should be used
- [X] Read the [docs](https://gql-tada.0no.co/).
- [X] Follow our [Code of Conduct](https://github.com/0no-co/gql.tada/blob/main/CODE_OF_CONDUCT.md)

JoviDeCroock:
Yeah, after the Vue compiler runs there's some weird wrapping on all the properties and we can't infer usage anymore. My best recommendation for now is to turn of field usage tracking in vue, maybe we should just turn it off for SFC 😅

JarvisH:
gql.tada is really cool, and using it seamlessly with Vue would be great. I know everything is still in an early phase so I completely understand if this is not a priority right now. Do you feel that long term complete Vue-Integration is possible?

kitten:
@JarvisH:
> only references inside <template> are ignored.

I think this is potentially not possible unless the transform always outputs the complete template code as TypeScript code as well.

[...]

PR #334, opened 2024-07-03
chore(cli): disable field-usage and colocated fragments for vue/svelte 

JoviDeCroock:
Resolves https://github.com/0no-co/gql.tada/issues/323

## Summary

The field-usage heuristics see the generated functions as two different scopes so the field-usage and co-located fragment tracking fails to report correctly, hence for now we disable this when analyzing svelte/vue.

### DIFF
--- a/.changeset/happy-cheetahs-provide.md
+++ b/.changeset/happy-cheetahs-provide.md
@@ -1,0 +1,5 @@
+---
+"@gql.tada/internal": minor
+---
+
+Add types for field-usage tracking and checking for co-located fragments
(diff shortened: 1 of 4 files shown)


.changeset/happy-cheetahs-provide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@gql.tada/internal": minor
```
kitten, line 2:
```suggestion
"@gql.tada/internal": patch
```

Seems to not be backwards-incompatible, if I'm not mistaken?

kitten (approved):

PR #340, opened 2024-07-09
fix(internal): Use TypeScript's built-in file watcher for SDL loader

### LINKED ISSUE #339: `gql.tada` not rebuilding on `schema.graphql` query return type changes
### Describe the bug

I found [this discussion](https://github.com/0no-co/gql.tada/discussions/328) mentioning that `graphql-env.d.ts` wasn't regenerating on schema change, but that the OP redacted their discussion.

However, I seemed to have found that when the _return type of a query_ is changed in the `schema.graphql`, then the .d.ts file isn't updated. However it seems to update fine when new properties are added or removed from the `schema.graphql`.

In order for me to see changes reflected in VScode, I have to CMD+P > Restart TS Server when query return types change. I have linked a sample repo.

Example of property being added/removed and LSP correctly catching it:

https://github.com/0no-co/gql.tada/assets/111302225/8c5d0f43-9a19-475d-b8ea-bb8b474f80c7

Example of changing the return type of a Query NOT updating the type until manual TS Server refresh:

https://github.com/0no-co/gql.tada/assets/111302225/b7a1501e-73ad-4dc8-bce4-bfbf71e7f9d7

### Reproduction

https://github.com/pmooney-socraticworks/gql-tada-issue

### gql.tada version

1.8.1

### Validations

kitten:
Resolves #339

## Summary

**TODO**
- Test this on Linux and Windows
- Check default options of `ts.sys.watchFile` that TypeScript itself uses (as default options were broken out of the box on macOS)

This switches the SDL loader over to using `ts.sys.watchFile` in favour of other, similar file watchers that work around some issues. Generally, there are some conditions under which the built-in file watcher in Node fails. It's suspected that swap-writing in some editors breaks the file watcher, for example.

Instead of working around this ourselves, we can use TypeScript's implementation, which implements all known workarounds and approaches, including polling fallbacks.

## Set of changes

- Use `ts.sys.watchFile` in SDL file loader's watcher

### DIFF
--- a/.changeset/silent-eyes-boil.md
+++ b/.changeset/silent-eyes-boil.md
@@ -1,0 +1,5 @@
+---
+"@gql.tada/internal": patch
+---
+
+Fix SDL-file schema watcher missing changed file events on macOS under certain write conditions.
--- a/packages/internal/src/loaders/sdl.ts
+++ b/packages/internal/src/loaders/sdl.ts
@@ -1,3 +1,4 @@
+import ts from 'typescript';
 import type { IntrospectionQuery } from 'graphql';
 import { buildSchema, buildClientSchema, executeSync } from 'graphql';
 import { CombinedError } from '@urql/core';
@@ -17,7 +18,7 @@ interface LoadFromSDLConfig {
 export function loadFromSDL(config: LoadFromSDLConfig): SchemaLoader {
   const subscriptions = new Set<OnSchemaUpdate>();
 
-  let controller: AbortController | null = null;
+  let abort: (() => void) | null = null;
   let result: SchemaLoaderResult | null = null;
 
   const load = async (): Promise<SchemaLoaderResult> => {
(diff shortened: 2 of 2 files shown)


JoviDeCroock (approved):

mrazauskas:
It seems that `ts.sys.watchFile` watchers are not unrefed. Opposite to `fs.watch` watcher that are created with `persistent: false`.

The unrefed watchers do not allow process to exit. That is fine for TypeScript language service, but that is troublesome in cases where `gql.tada` is used as stand alone package.

---

I bumped into this limitation while debugging this issue: https://github.com/tstyche/tstyche/pull/276

Reproduction repo can be found here: https://github.com/tstyche/tstyche/pull/276#issuecomment-2267647218

Not sure what could be the solution. Just wanted to draw your attention.

mrazauskas:
> Generally, there are some conditions under which the built-in file watcher in Node fails. It's suspected that swap-writing in some editors breaks the file watcher, for example.

@kitten The other difference previous implementation was using the async version of `fs.watch`. In contrary TypeScript's watchers are wrapped around sync `fs.watch`. I was experimenting with both of them while implementing watch mode for TSTyche. Indeed sometimes events were missing in the async `fs.watch`.

I was wondering if you tried using the sync `fs.watch` implementation instead?

PR #342, opened 2024-07-10
docs: add vite-specific oddity

DerTimonius:
## Summary

First of all, thanks for this awesome tool!

This PR will add a small section to the installation docs: while installing this in a Vite project, I realized after some trial and error that it is also necessary to add the plugin to the `tsconfig.app.json` file, as you won't get type hints otherwise (yes, I have spent way too long to figure that out).

## Set of changes

Just small changes to the docs

### DIFF
--- a/website/get-started/installation.md
+++ b/website/get-started/installation.md
@@ -62,6 +62,10 @@ language server. This is the main configuration for both the TypeScript plugin a
 Setting up `gql.tada/ts-plugin` will start up a [“TypeScript Language Service Plugin”](https://github.com/microsoft/TypeScript/wiki/Writing-a-Language-Service-Plugin#whats-a-language-service-plugin) when TypeScript is analyzing a file in our IDE or editor. This provides editor hints, such as diagnostics,
 auto-completions, and type hovers for GraphQL.
 
+> [!NOTE] Vite Setup
+> If you are working in a Vite project, it is also necessary to add the Typescript plugin to the `tsconfig.app.json` file.
+> Otherwise you won't get any type hints while writing queries.
+
 > [!NOTE] VSCode Setup
 > There may be extra steps you should take when you're using VSCode.
 > [Read about these steps in the "VSCode Setup" section below.](#vscode-settings-and-plugins)


kitten:
Can you clarify your setup please?

This really seems like an oddity of your own project. For instance, take our examples in this repo. They're pretty standard Vite projects and they don't contain a `tsconfig.[name].json` at all, but only a single `tsconfig.json`: https://github.com/0no-co/gql.tada/tree/main/examples/example-pokemon-api

Furthermore, by default renaming the `tsconfig.json` means that they won't be picked up by your editor, so out of the box, this wouldn't actually work unless your main `tsconfig.json` refers to the other `tsconfig.app.json`, e.g. with an `extends`.

I'm more inclined to ask maybe what `gql.tada doctor` said for you in this case. If the output here doesn't warn you about this then we can probably work on that, but as it is, the docs addition doesn't quite make sense to me. Sorry! ❤️

DerTimonius:
I actually did nothing more than run `pnpm create vite` and it created three different tsconfig files: 
- `tsconfig.json`
- `tsconfig.node.json`
- `tsconfig.app.json`

when I ran `gql.tada doctor` it didn't give me any errors, and the creation of the correct types of the query worked, but there were no type hints.

only after I put the plugin section into the `tsconfig.app.json` file, the type hints were actually there.

DerTimonius:
Hm..I did not think of the differences in the created tsconfigs by Vite (insert relevant xkcd about standards here). 

Yes, the initial `tsconfig.json` looked just like the file you linked.

I agree that checking for referenced tsconfig files makes sense. The problem is not only `gql.tada doctor`, but also other `gql.tada` CLI tools like `check` or `generate output` when I would remove the `compilerOptions` field from the `tsconfig.json` and only keep it in `tsconfig.app.json`.

So in my opinion, there are two different issues with Vite as a user: 
1. when following the instructions in the docs (or running `gql.tada init`), the package works without type hints
2. if there are multiple tsconfig files, you might end up putting the `compilerOptions` into the wrong file and the CLI won't work

Therefore there are two different solutions, with the first one being easier:
1. add more generic instructions to the docs on what to do if you have multiple `tsconfig.json` files (in this case, just add the compilerOptions twice)
2. add checks to the CLI, if there are references to other tsconfig files

What do you think? What approach do you prefer?

martywallace:
Just spent a couple hours pulling my hair out over intellisense not working in the `graphql(...)` block and randomly stumbled upon this MR which solved my issue (moved the plugin from `tsconfig.json` to `tsconfig.app.json`), so even if it doesn't get merged, thank you!

PR #349, opened 2024-07-29
patch: Match optional (`T | {}`) fragment data

kitten:
## Summary

Previously, when a fragment as annotated to be optional with a directive (such as `@skip`, `@include`, or `@defer`), `readFragment` wasn't able to match such a type. Optional fragments are typed as `T | {}` since they may be optional, however, no match pattern in `readFragment` was able to match this pattern.

The patterns have now been extended to match these cases and pass them through as `T | {}` on the return type.

## Set of changes

- Refactor inferred `readFragment` pattern to add `T` type to match all patterns in fewer overloads
- Extend explicit-generic-overloads to account for `| {}` explicitly (doubling them)

### DIFF
--- a/.changeset/quiet-panthers-wink.md
+++ b/.changeset/quiet-panthers-wink.md
@@ -1,0 +1,5 @@
+---
+"gql.tada": patch
+---
+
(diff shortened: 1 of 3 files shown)


website/get-started/installation.md
```diff
@@ -5,42 +5,45 @@ description: How to get set up and ready
 
 # Installation
 
-We’ll go through the steps to get `gql.tada` set up properly.
+The `gql.tada` package provides typings and the runtime API as a library,
```
JoviDeCroock, line 8: Seems unrelated as well

JoviDeCroock (approved):

PR #352, opened 2024-07-29
feat(examples): use new plugin alias

leomelzer:
## Summary

Thank you for making gql.tada! It feels like a breath of fresh air after many years of the good and trusted graphql-codegen. 🎉 

When browsing the source code (especially the examples) I was a little confused:

- The examples are not using the newish `gql-tada/ts-plugin` alias. Not sure if this is intended.
- The example for React contained the somewhat (I think) advanced concept Persisted Queries but no simple showcase of variables and directives.

Hence...

## Set of changes

- all examples: removes `@0no-co/graphqlsp` dependency and instead uses the new plugin alias
- example-pokemon-api: Make the example a little interactive with variables and a directive

I can update the Vue + Svelte Examples to match the React one if there is interest.

Thank you!

### DIFF
--- a/.changeset/lemon-rice-arrive.md
+++ b/.changeset/lemon-rice-arrive.md
@@ -1,0 +1,5 @@
+---
+"gql.tada": minor
+---
+
+Use gql.tada/ts-plugin alias for examples
--- a/examples/example-pokemon-api/package.json
+++ b/examples/example-pokemon-api/package.json
@@ -13,7 +13,6 @@
     "urql": "^4.0.7"
   },
   "devDependencies": {
-    "@0no-co/graphqlsp": "^1.12.9",
     "@types/react": "^18.2.79",
     "@types/react-dom": "^18.2.25",
     "@vitejs/plugin-react": "^4.2.1",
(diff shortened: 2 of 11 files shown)


kitten:
(side-note, a changeset isn't needed for non-user-facing changes, since the library itself isn't going to change and `examples/` aren't published or subject to a changelog)

I've deliberately had to leave this out. When I implemented the alias, I noticed that TypeScript plugins aren't being correctly resolved in monorepos where the plugin is aliased from a different folder. In this case, `gql.tada` is linked via a symlink to basically `../../` relative to the examples' folders. TypeScript however seems to confuse the root folder used for this symlink and eventually attempts to then load the plugin from `<rootDir>/../../` as far as I remember. This completely breaks the plugin.

I haven't gotten around to reporting and tracking this back down in TypeScript itself yet, but even if we do, this would mean that we couldn't change the examples for quite a while.
Since we still have (and will likely keep) the instructions here, I skipped the change to `examples` and didn't update them: https://gql-tada.0no.co/get-started/installation#prior-to-typescript-5-5

leomelzer:
Thanks for the quick feedback @kitten. 

Got it, I saw the behavior, a little tricky...I was first looking at the example in isolation with `npm install` where everything was working fine. When I was properly setting up the monorepo and running the checks (`pnpm multi run check`) I noticed it was broken. 

Unfortunately I lack a little context for reporting that in the TS repo myself. 

If you find [33e1987](https://github.com/0no-co/gql.tada/pull/352/commits/33e1987e178c928109237b84b5a1906ec9157a70) interesting I can open that as a separate PR. In any case I'll go ahead and close here. Thanks!

PR #353, opened 2024-07-30
fix(cli-utils): Support `@vue/language-core@2.0.28` changes

kitten:
Similar past change: https://github.com/0no-co/gql.tada/pull/305

## Summary

This supports changes in `@vue/language-core` which renamed `getBasePlugins` (again) to `createPlugins`, which caused Vue files to again not be parsed.

To prevent further delays in debugging such errors, there's now a `check` function that outputs an explicit error if `VueVirtualCode` or `createPlugins` are undefined, which is a pretty sure signal of breaking changes.

## Set of changes

- Upgrade internal test version of `@vue/language-core`
- Fix `@vue/language-core` version, which was accidentally bumped (set to `~2.0.0`)
- Add `check` function with special error for breaking Vue changes

### DIFF
--- a/.changeset/shaggy-rivers-move.md
+++ b/.changeset/shaggy-rivers-move.md
@@ -1,0 +1,5 @@
+---
+"@gql.tada/cli-utils": patch
+---
+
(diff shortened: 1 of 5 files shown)


JoviDeCroock (approved):
.... Why....

Great find bud

PR #355, opened 2024-07-30
patch(cli-utils): Bump stack trace limit when TypeScript is used

kitten:
## Summary

Increase stack trace limit to `25` when TypeScript is used, as the default of `10` isn't enough to properly debug issues in most (if not all) cases that TypeScript's type checker or program is involved in a crash.

## Set of changes

- Add `bumpStackTraceLimit` utility

### DIFF
--- a/packages/cli-utils/src/ts/factory.ts
+++ b/packages/cli-utils/src/ts/factory.ts
@@ -44,6 +44,13 @@ export interface ProgramFactory {
   build(): ProgramContainer;
 }
 
+/** Bumps the Error stack traces to a length of 100 for better debugging. */
+const bumpStackTraceLimit = () => {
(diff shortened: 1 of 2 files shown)


packages/cli-utils/src/ts/factory.ts
```diff
@@ -44,6 +44,13 @@ export interface ProgramFactory {
   build(): ProgramContainer;
 }
 
+/** Bumps the Error stack traces to a length of 100 for better debugging. */
```
JoviDeCroock, line 47:
```suggestion
/** Bumps the Error stack traces to a length of 25 for better debugging. */
```

JoviDeCroock (approved):

PR #386, opened 2024-08-29
fix: handle multiple tsconfig `extends` for resolving GraphQLSP plugin entry

aweber1:
## Summary

Given the following `tsconfig.json` setup:

```
// tsconfig.json
{
  "extends": ["./tsconfig.app.json", "./tsconfig.gql-tada.json"]
}

// tsconfig.app.json
{
  "compilerOptions": {
     // some options here
   }
}

// tsconfig.gql-tada.json
{
  "compilerOptions": {
    "plugins": [
      {
        "name": "gql.tada/ts-plugin",
        // schema config
      }
    ]
  }
}
```

I encounter the following error:
`Could not find a valid GraphQLSP plugin entry in: tsconfig.json`

The code in this PR fixes the issue for me, but there aren't unit tests specifically for this area of code so not sure of any adverse effects.

## Set of changes

The change was pretty straightforward in just modifying `return load(tsconfigPath)` to be `return await load(tsconfigPath)` so that exceptions thrown when trying to load nested tsconfig "extends" elements are caught (and swallowed) by the intended `catch(_error)` statements.

### DIFF
--- a/.changeset/fluffy-seahorses-flow.md
+++ b/.changeset/fluffy-seahorses-flow.md
@@ -1,0 +1,5 @@
+---
+'@gql.tada/internal': patch
+---
+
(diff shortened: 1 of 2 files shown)


kitten (approved):
Nice catch! didn't spot this typo when refactoring this file. I think this must've regressed when this was made more reusable

Issue #446, opened 2025-03-22
`gql-tada doctor` says version is out of date when using a pnpm workspace catalog


rik-iso:
### Describe the bug

Using [pnpm workspace catalogs](https://pnpm.io/catalogs) in a pnpm monoreo, the doctor complains my version is out of date:

```
The version of gql.tada in your dependencies is out of date.
┗ ℹ It's recommended to upgrade gql.tada to at least 1.0.0
```

In reality, it's set to `catalog:`, and in pnpm-workspace.yaml I have an up to date version.

This is very new pnpm syntax. I'd be happy to submit a PR fixing it - just wanted to check it's not on your roadmap or a will-not-fix.

### gql.tada version

latest

### Validations

- [x] I can confirm that this is a bug report, and not a feature request, RFC, question, or discussion, for which [GitHub Discussions](https://github.com/0no-co/gql.tada/discussions) should be used
- [x] Read the [docs](https://gql-tada.0no.co/).
- [x] Follow our [Code of Conduct](https://github.com/0no-co/gql.tada/blob/main/CODE_OF_CONDUCT.md)

kitten:
The logic for this is in this file: https://github.com/0no-co/gql.tada/blob/5b5d1e49e32b0e80f6ccd21820991be8363de845/packages/cli-utils/src/commands/doctor/helpers/versions.ts

I simply didn't get to this, and since it's still a bit of a niche feature, I didn't bother adding it. Sorry! Happy to solve this however you see fit, but also happy for you to file a PR for it of course

PR #460, opened 2025-06-28
Add support for vitest workspace catalogs

JoviDeCroock:
Resolves https://github.com/0no-co/gql.tada/issues/446

## Summary

This adds support for `pnpm workspace catalog`, this basically entails that a version can be specified as `catalog:` which forces us to look it up in the `pnpm-workspace.yaml` file.

I am working on tests but I forgot how tedious it is to make these things work on Windows devices 😅 Yikes to the test timeout in CI, this is due to the async import of typescript just taking long.

### DIFF
--- a/.changeset/modern-walls-relax.md
+++ b/.changeset/modern-walls-relax.md
@@ -1,0 +1,5 @@
+---
+'@gql.tada/cli-utils': minor
+---
+
+Add support for pnpm workspace catalog version resolution to `gql-tada doctor`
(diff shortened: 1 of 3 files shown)


kitten (approved):
This might be more effort than what it's worth but other package managers are adopting this, so we could have a quick spot-check if it's correct for others: https://bun.sh/docs/install/catalogs

That said, `doctor` of course doesn't aim for 100% correctness, so this is fine either way 😅

PR #473, opened 2025-07-21
Fix external scalars 2

JoviDeCroock:
## What's the problem?

When generating the turbo cache, we collect all the `graphql()` call expressions and their types, but we weren't copying over the imports from the GraphQL source files (like `examples/example-pokemon-api/src/graphql.ts`). This meant the generated cache file was missing imports like:

```ts
import type { introspection } from './graphql-env.d.ts';
import type { Test } from './Test';
```

## What's the fix?

- **Trace graphql() calls back to source**: Added AST traversal to find where each `graphql()` call is imported from
- **Collect imports**: Extract all import statements from GraphQL source files 
- **Adjust paths**: Update relative import paths to work from the turbo cache location
- **Avoid circular imports**: Filter out any imports that would reference the output files themselves
- **Deduplicate**: Only process each unique GraphQL source file once

## Example

Before:
```ts
// turbo cache only had:
import type { TadaDocumentNode, $tada } from 'gql.tada';
```

After:
```ts  
// turbo cache now includes the GraphQL source imports:
import type { TadaDocumentNode, $tada } from 'gql.tada';
import type { Test } from '../examples/example-pokemon-api/src/Test';
```

Works for both single and multi-schema setups.

### DIFF
--- a/packages/cli-utils/src/commands/turbo/runner.ts
+++ b/packages/cli-utils/src/commands/turbo/runner.ts
@@ -33,11 +33,49 @@ export async function* run(tty: TTY, opts: TurboOptions): AsyncIterable<ComposeI
   } catch (error) {
     throw logger.externalError('Failed to load configuration.', error);
   }
+  const projectPath = path.dirname(configResult.configPath);
+
+  let destination: WriteTarget;
+  if ('schema' in pluginConfig) {
+    if (!opts.output && tty.pipeTo) {
+      destination = tty.pipeTo;
+    } else if (opts.output) {
+      destination = path.resolve(process.cwd(), opts.output);
+    } else if (pluginConfig.tadaTurboLocation) {
+      destination = path.resolve(projectPath, pluginConfig.tadaTurboLocation);
+    } else if (pluginConfig.tadaOutputLocation) {
+      destination = path.resolve(
+        projectPath,
+        pluginConfig.tadaOutputLocation,
+        '..',
+        'graphql-cache.d.ts'
+      );
+      yield logger.hintMessage(
+        'No output location was specified.\n' +
+          `The turbo cache will by default be saved as ${logger.code('"graphql-cache.d.ts"')}.\n` +
+          logger.hint(
+            `To change this, add a ${logger.code('"tadaTurboLocation"')} in your configuration,\n` +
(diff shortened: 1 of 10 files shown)


packages/cli-utils/src/commands/turbo/runner.ts
```diff
@@ -34,6 +35,16 @@ export async function* run(tty: TTY, opts: TurboOptions): AsyncIterable<ComposeI
...
 
+  // Read the TypeScript configuration
+  const tsConfig = await readTSConfigFile(configResult.configPath);
+
+  // Create TypeScript compiler options for proper module resolution
+  const compilerOptions = ts.parseJsonConfigFileContent(
+    tsConfig || {},
+    ts.sys,
+    configResult.rootPath
+  ).options;
```
kitten, line 46: I don't think this needs to end up here, given that `runTurbo` already does resolution, so this could be moved there, which also seems more isolated. The problem with having it here isn't just the thread isolation, but here, we also have to recreate TS harness stuff from scratch, which seems very ad-hoc and error-prone, especially since it's potentially not as aware of modifications (we don't use/create the compiler options like that, and this changes the `moduleResolution` override)

In `thread` we also have the container: https://github.com/0no-co/gql.tada/blob/a37ca0bf527520e4e6b879b74794323253471308/packages/cli-utils/src/commands/turbo/thread.ts#L187C1-L187C37

There, the resolution already does half of the remapping dne here, and we do have access to the `rootPath`, and can pass an `outputPath`, potentially for relative resolution: https://github.com/0no-co/gql.tada/blob/a37ca0bf527520e4e6b879b74794323253471308/packages/cli-utils/src/commands/turbo/thread.ts#L77C23-L77C40

We could also bypass `createCompilerHost` and don't have to call it directly, as that's done here already and we can expose it: `createVirtualCompilerHost` from: https://github.com/0no-co/gql.tada/blob/a37ca0bf527520e4e6b879b74794323253471308/packages/cli-utils/src/ts/factory.ts#L77

We can either expose it or add a method for resolution there, but this ensures that we take the virtual files into account, if necessary, so that I think will prevent us from being able to forget about that part of the TS vfs stuff in the future 🤔

PR #479, opened 2025-07-24
Support parsing executable definition descriptions

JoviDeCroock:
Similar to https://github.com/0no-co/graphql.web/pull/59 but in the TS parser, making sure this doesn't start failing. For the LSP to support this we'll need to upgrade GraphQL.web

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -36,6 +36,12 @@ export type takeValue<In extends any[], Const extends boolean> =
       : void
     : void;
 
+export type takeString<In extends any[]> = In extends [Token.String, ...infer In]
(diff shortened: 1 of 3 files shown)


src/parser.ts
```diff
@@ -277,7 +277,10 @@ type _takeVarDefinitionRec<Definitions extends any[], In extends any[]> = In ext
   ? _match<Definitions, In>
   : takeVarDefinition<In> extends _match<infer Definition, infer In>
     ? _takeVarDefinitionRec<[...Definitions, Definition], In>
-    : void;
+    : takeValue<In, true> extends _match<infer _, infer In>
```
kitten, line 280: We should be able to split out a new `takeString` here without any performance consequences probably

PR #481, opened 2025-08-05
Add support for nodenext

### LINKED ISSUE #480: graphql-cache.d.ts has invalid imports when using moduleResolution nodenext
### Describe the bug

My tsconfig.json has moduleResolution 'nodenext' which forces the use of .js or .ts extension in order to resolve imports. 

The `graphql-cache.d.ts` file gets generated with imports that end in `gql.tada/dist/gql-tada` for me which results in the imported type being `any` for `TypedDocumentNode`. If I change these to end in either ".js" or ".ts" I get the correct type.

Any way to either honor this moduleResolution setting when generating the import line or give an option to simply force extensions on imports?

(note: graphql codegen has the config option: `emitLegacyCommonJSImports: false`, to work around this. )

### Reproduction

_No response_

### gql.tada version

gql.tada@1.8.12

### Validations

JoviDeCroock:
Resolves #480 

~~We should not merge this right away as we need to investigate the why of this original default and whether it's still needed~~

We should revisit the Bundler fallback in the next major, for the time being I've found a backwards compat way.

### DIFF
--- a/packages/cli-utils/src/commands/turbo/thread.ts
+++ b/packages/cli-utils/src/commands/turbo/thread.ts
@@ -89,7 +89,8 @@ function collectImportsFromSourceFile(
   sourceFile: ts.SourceFile,
   pluginConfig: GraphQLSPConfig,
   resolveModuleName: (importSpecifier: string, fromPath: string, toPath: string) => string,
-  turboOutputPath?: string
+  turboOutputPath?: string,
+  compilerOptions?: ts.CompilerOptions
 ): GraphQLSourceImport[] {
   const imports: GraphQLSourceImport[] = [];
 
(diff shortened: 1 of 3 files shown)


packages/cli-utils/src/commands/turbo/thread.ts
```diff
@@ -103,13 +104,22 @@ function collectImportsFromSourceFile(
...
-            .replace(/\.ts$/, '')
-            .replace(/\.tsx$/, '');
+          );
+
+          if (shouldTreatImportsAsNodeNext) {
+            if (adjustedSpecifier.endsWith('.ts') || adjustedSpecifier.endsWith('.tsx')) {
+              adjustedSpecifier = adjustedSpecifier
+                .replace(/\.ts$/, '.js')
+                .replace(/\.tsx$/, '.js');
+            }
```
kitten, line 118: I'm a bit concerned about this, but I can investigate a replacement separately by looking at what other implementors of TS on-the-fly transpilation do

kitten (approved):

packages/cli-utils/src/commands/turbo/thread.ts
kitten: afaik, there's no guarantee that the file will be resolved as such due to cjs/mjs extensions. Mainly, we're resolving a type from a file path, so in theory TypeScript will have some information (or canonical API) to resolve the file in its loaded sources without us essentially having to match/guess the filepath. Hence, I'm a little concerned that it feels like we're trying to match or guess TS' required resolution behaviour here

PR #524, opened 2026-05-21
Isolate monorepo packages

aarne:
## Summary

Fixes a performance regression in monorepo setups where `gql-tada check` scans the entire workspace instead of just the target project.

**Root cause:** When the `@0no-co/graphqlsp` plugin is declared in a shared root tsconfig (e.g. tsconfig.base.json) and individual projects extend it, `loadConfig` returns `configPath` pointing to the root tsconfig while `rootPath` is the project directory. `resolveConfig` was parsing the root config unconditionally, so TypeScript resolved `baseUrl: "."` against the workspace root — causing `programFactory` to walk every `.ts` file in the entire monorepo on every `check` invocation.

**Fix:** In `resolveConfig`, when `configPath` lives in a different directory than `rootPath`, look for a `tsconfig.json` directly in `rootPath` and use that instead. The project's own tsconfig carries the correct `include` list and relative paths, scoping the TypeScript program to just the project. 

## Set of changes

- `resolveConfig`: detect when `configPath` (parent tsconfig) is in a different directory from `rootPath` (project dir) and fall back to the project's own `tsconfig.json` when one exists
- test that creates a tmp monorepo with a shared tsconfig.base.json at the root and a project with its own scoped `tsconfig.json`, asserts that `programFactory.projectDirectories` contains only paths inside the project

## AI Disclosure

Claude may have been involved :) Especially in digging through and identifying the root cause of the slowdown.

### DIFF
--- a/packages/cli-utils/src/ts/factory.ts
+++ b/packages/cli-utils/src/ts/factory.ts
@@ -272,15 +272,31 @@ const resolveDefaultLibsPath = (params: ProgramFactoryParams): string => {
 };
 
 const resolveConfig = (params: ProgramFactoryParams, system: ts.System): ts.ParsedCommandLine => {
-  const text = system.readFile(params.configPath, 'utf8') || '{}';
-  const parseResult = ts.parseConfigFileTextToJson(params.configPath, text);
+  let configFile = params.configPath;
+  const basePath = params.rootPath || path.dirname(params.configPath);
+
+  // When the plugin config is found in a parent tsconfig (via `extends`), `loadConfig` returns
+  // configPath pointing to the parent (e.g. tsconfig.base.json at workspace root) while rootPath
+  // is the project directory. Parsing the parent config uses its directory as context, causing
+  // TypeScript to resolve `baseUrl: "."` against the workspace root and scan the entire monorepo.
+  // Fix: when configPath is in a different directory than rootPath, use the project's own
+  // tsconfig.json instead, which carries the correct `include` list and relative `baseUrl`.
+  // We use ts.sys (real TS sys) because the VFS blocks all tsconfig.json fileExists lookups.
+  if (params.rootPath && path.dirname(params.configPath) !== params.rootPath) {
+    const projectTsconfig = path.join(params.rootPath, 'tsconfig.json');
+    if (ts.sys.fileExists(projectTsconfig)) {
+      configFile = projectTsconfig;
+    }
+  }
+
+  const text = system.readFile(configFile, 'utf8') || '{}';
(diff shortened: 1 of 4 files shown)


packages/cli-utils/src/ts/factory.ts
```diff
@@ -272,15 +272,31 @@ const resolveDefaultLibsPath = (params: ProgramFactoryParams): string => {
...
+
+  // When the plugin config is found in a parent tsconfig (via `extends`), `loadConfig` returns
+  // configPath pointing to the parent (e.g. tsconfig.base.json at workspace root) while rootPath
+  // is the project directory. Parsing the parent config uses its directory as context, causing
+  // TypeScript to resolve `baseUrl: "."` against the workspace root and scan the entire monorepo.
+  // Fix: when configPath is in a different directory than rootPath, use the project's own
+  // tsconfig.json instead, which carries the correct `include` list and relative `baseUrl`.
+  // We use ts.sys (real TS sys) because the VFS blocks all tsconfig.json fileExists lookups.
+  if (params.rootPath && path.dirname(params.configPath) !== params.rootPath) {
+    const projectTsconfig = path.join(params.rootPath, 'tsconfig.json');
```
JoviDeCroock, line 286: Can we avoid hard-coding `rootPath/tsconfig.json` here? `gql.tada check --tsconfig tsconfig.app.json` is valid, but `loadConfig` only gives us the parent plugin config plus the directory, so we’ve lost the original target config filename and may parse the wrong file.

vitest.config.ts
```diff
@@ -1,6 +1,26 @@
...
+// checkout (no dist/, not symlinked). Any test that loads a module chain
+// ending at these packages needs them to be resolvable, even if the actual
+// lazy-import code path is never executed. This plugin creates lightweight
+// virtual stubs so Vite never hits the "Failed to resolve entry" error.
+function stubOptionalPeerPlugin(): Plugin {
+  const stubs = new Set(['@gql.tada/svelte-support', '@gql.tada/vue-support']);
+  return {
+    name: 'vitest:stub-optional-peers',
+    resolveId(id) {
+      if (stubs.has(id)) return '\0stub:' + id;
```
JoviDeCroock, line 14: This still fails for me from a clean checkout because CI runs tests before the workspace build. We might just need a Vitest alias for `@gql.tada/svelte-support` / `@gql.tada/vue-support` to point at source instead of dist; that’d also be less misleading than stubbing the modules.

JoviDeCroock (commented):
Left a couple notes.

kitten:
Superseded by #545

PR #525, opened 2026-05-29
Update vue-support to support language-core 3.x

KammererTob:
## Summary

`@gql.tada/vue-support` declares `@vue/language-core: ~2.0.0`. The ~2.0.0 semver range allows only >=2.0.0 <2.1.0. 

Projects using `vue-tsc@3.x` have `@vue/language-core@3.x` at the root, so npm/yarn creates a nested install of `@vue/language-core@2.0.29` inside `node_modules/@gql.tada/vue-support/node_modules/`. 
That nested version brings in `@vue/compiler-core@3.5.24`, which depends on entities@^4.5.0. 
But `@vue/compiler-core` recently has updated the version of entities to 7.0.1, which restructured its exports and dropped ./lib/decode.js. 

This leads to `gql.tada turbo` to fail with this error:
```
 ⚠ System Error 
Could not build cache
┗ For Vue support the `@gql.tada/vue-support` package must be installed.
  Install the package and try again.
```

## Set of changes

- Allow `@vue/language-core": "^2.0.0 || ^3.0.0"` (so also 3.x)
- Change how compiler options are resolved to work with both language-core 2.x and 3.x (later versions of 2.x also exported a `getDefaultCompilerOptions`, but earlier ones did not)

### DIFF
--- a/.changeset/calm-tigers-jog.md
+++ b/.changeset/calm-tigers-jog.md
@@ -1,7 +1,5 @@
 ---
 '@gql.tada/vue-support': patch
-'gql.tada': patch
-'@gql.tada/cli-utils': patch
 ---
 
 Support @vue/language-core version 3.0.0. This fixes compatibility issues with projects having a dependency on vue-tsc@3.x (through the transitive dependency of entities, which changed it's exports)
--- a/.changeset/calm-tigers-jog.md
+++ b/.changeset/calm-tigers-jog.md
@@ -1,0 +1,7 @@
+---
+'@gql.tada/vue-support': patch
+'gql.tada': patch
+'@gql.tada/cli-utils': patch
+---
+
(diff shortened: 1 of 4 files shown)


.changeset/calm-tigers-jog.md
```diff
@@ -0,0 +1,7 @@
+---
+'@gql.tada/vue-support': patch
+'gql.tada': patch
+'@gql.tada/cli-utils': patch
```
JoviDeCroock:
```suggestion
```

JoviDeCroock:
CI is failing due to an outdated lockfile

KammererTob:
Fixed. Sorry about that.

JoviDeCroock (approved):

PR #529, opened 2026-06-01
fix: surface turbo worker failures

### LINKED ISSUE #503: `gql-tada turbo` silently fails when running out of memory
### Describe the bug

I had a CI check to make sure that generated types that are committed are up to date. The job was passing, even when I knew the types were outdated. After much debugging and exploration, I realized that it was a memory issue. This was proven by the fact that a larger GitHub runner with 32GB of RAM addressed the issue. When running `gql-tada turbo` it checks more than 1,500 files so it must be running into some sort of limit.

### Reproduction

I don't have a reproduction available since this was done in a private repo with a large amount of files. Perhaps the maintainers could identify the code path that could cause this issue. If a reproduction is absolutely required I could look into creating a fake repo that reproduces the issue.

### gql.tada version

gql.tada@1.8.10

### Validations

JoviDeCroock:
## Summary
- keep turbo worker threads referenced while work is active
- reject pending turbo pulls on worker errors and unexpected exits
- add regression coverage for worker lifecycle/error propagation

Closes #503

### DIFF
--- a/.changeset/silent-workers-doubt.md
+++ b/.changeset/silent-workers-doubt.md
@@ -1,0 +1,5 @@
+---
+'@gql.tada/cli-utils': patch
+---
+
+Surface worker thread errors and unexpected exits during turbo generation instead of allowing pending work to resolve silently.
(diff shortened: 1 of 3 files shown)


kitten (approved):
Tested on a repo that OOMs and can confirm the error surfaces as expected

PR #538, opened 2026-06-12
feat: Support tsconfig references when resolving plugin configurations

### LINKED ISSUE #452: RFC: Search in tsconfig.references too
## Summary

When searching for plugin config in tsconfig.json, gql.tada should also check the files listed in `tsconfig.references` too.

I use Vite with their [react-ts](https://stackblitz.com/edit/vitejs-vite-dufwsnja?file=package.json&terminal=dev) template. The template contains multiple tsconfig files:
- tsconfig.node.json - used exclusively for `vite.config.ts`
- tsconfig.app.json - used for everything else in `src` folder
- tsconfig.json - has "references" field listing the files above, and contains no other options

With this setup, the only way to get something out of gql.tada is to place plugin config into `tsconfig.app.json` and run `gql-tada check --tsconfig tsconfig.app.json`. Placing plugin config in any other files, including `tsconfig.json`, and/or omitting `--tsconfig` option results in either "No problems found" (because no files were checked), or in failing to run because no plugin config found.

## Proposed Solution

`load(targetPath)` function from `@gql.tada/internal/src/resolve.ts`, apart from checking `tsconfig.extends` field, should also check `tsconfig.references` and try to load plugin config from there.

I forked gql.tada repo and [implemented the change](https://github.com/0no-co/gql.tada/compare/main...mrsimb:gql.tada:load-tsconfig-references), but i don't know how to properly link the local version of the package to my project to test it. I can open a PR if you're ok the changes i made.

JoviDeCroock:
## Summary

Resolves #452.

Vite/Vue/Nuxt templates ship a solution-style root `tsconfig.json` that only contains `references` to `tsconfig.app.json`/`tsconfig.node.json` (and `files: []`). The CLI previously only followed `extends` when searching for the GraphQLSP plugin entry, so it either errored with "Could not find a valid GraphQLSP plugin entry" or checked zero files. As discussed in the issue, a "first reference wins" lookup isn't enough since references are parallel projects — in monorepos, multiple referenced projects can each have their own gql.tada plugin configuration, and all of them should be active.

## What's implemented

**`@gql.tada/internal`**
- New `loadConfigs(targetPath?)` resolves the root tsconfig plus all transitively referenced projects that carry a plugin entry (each project still follows its own `extends` chain). Reference paths resolve relative to the referencing config (directory references get `tsconfig.json` appended, mirroring `ts.resolveProjectReferencePath`), traversal is cycle-safe, and unreadable references are skipped like unresolvable `extends`.
- Solution-style configs (`references` + `files: []`, no `include`) carrying a plugin entry are only used as a fallback when no referenced project has one, matching tsserver semantics and keeping the previously documented "plugin in the root tsconfig" workaround working.
- `LoadConfigResult` gains `tsconfigPath` (the project's own config file, as opposed to `configPath`, which may be an `extends` base). `loadConfig` returns the first `loadConfigs` result, so GraphQLSP's `resolveTypeScriptRootDir` keeps working unchanged.
- New `validateUniqueOutputLocations` throws a `TadaError` naming both projects when two distinct projects resolve `tadaOutputLocation`, `tadaTurboLocation`, or `tadaPersistedLocation` to the same file, mirroring the multi-schema uniqueness rules.

**`@gql.tada/cli-utils`** — a shared `loadProjects()` helper feeds every command:
- `check` runs diagnostics per project and aggregates one severity summary, so `--fail-on-warn`/`--min-severity` and GitHub annotations behave as before.
- `turbo` and `generate-persisted` write per-project outputs; with `--fail-on-warn`, a warning project's output is skipped but remaining projects still run, failing at the end with the aggregate count. An explicit `--output` (or piping) errors when multiple projects are configured.
- `generate-output` loops projects in-process; `generate-schema` asks for `--output` when the destination is ambiguous across projects.
- `doctor` validates every project's plugin config, reports conflicts, and loads every project's schema.
- `init` now writes the plugin entry into `tsconfig.app.json` (or the first readable reference) when the root tsconfig is solution-style, instead of producing the broken setup from the issue.

## Behavior notes (flagged in the changeset)

- Threads now derive each project's file list from the project's own tsconfig (`parseJsonConfigFileContent` merges `extends` itself) instead of the `extends` base file the plugin entry was found in, and relative output locations resolve from the project's tsconfig directory.
- `extends` is resolved relative to each config file rather than the root config's directory.

## Testing

- New fixture-based suites: `packages/internal/src/__tests__/resolve.test.ts` (12 cases: Vite shape, monorepo, extends, shared base, diamond/cycle, directory + non-standard references, missing references, solution fallback), `config.test.ts` (conflict validation), and `packages/cli-utils/src/commands/shared/__tests__/projects.test.ts`, which runs `generate-output` end-to-end against a Vite-shaped fixture and a two-package monorepo fixture and asserts the written `graphql-env.d.ts` files.
- `tsc --noEmit` and `eslint` pass for both packages.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/internal/src/__tests__/resolve.test.ts
+++ b/packages/internal/src/__tests__/resolve.test.ts
@@ -10,7 +10,7 @@ const PLUGIN = { name: '@0no-co/graphqlsp', schema: './schema.graphql' };
 type Fixture = Record<string, unknown>;
 
 const inFixture = async (files: Fixture, fn: (root: string) => Promise<void>): Promise<void> => {
-  const root = await fs.mkdtemp(path.join(os.tmpdir(), 'gql-tada-resolve-'));
+  const root = await fs.realpath(await fs.mkdtemp(path.join(os.tmpdir(), 'gql-tada-resolve-')));
   try {
     for (const name of Object.keys(files)) {
       const filePath = path.join(root, name);
--- a/packages/internal/src/__tests__/resolve.test.ts
+++ b/packages/internal/src/__tests__/resolve.test.ts
@@ -194,6 +194,21 @@ describe('loadConfigs', () => {
     );
   });
 
+  it('rejects malformed referenced config files', () => {
+    return inFixture(
+      {
+        'tsconfig.json': {
+          files: [],
+          references: [{ path: './tsconfig.app.json' }],
+        },
+        'tsconfig.app.json': '{',
+      },
+      async (root) => {
+        await expect(loadConfigs(root)).rejects.toThrow(/tsconfig\.app\.json/);
+      }
+    );
+  });
+
   it('resolves a non-solution root config with plugin-less references', () => {
     return inFixture(
       {
--- a/packages/internal/src/__tests__/resolve.test.ts
+++ b/packages/internal/src/__tests__/resolve.test.ts
@@ -1,0 +1,266 @@
+import * as os from 'node:os';
+import * as path from 'node:path';
+import * as fs from 'node:fs/promises';
+import { describe, it, expect } from 'vitest';
+
+import { loadConfig, loadConfigs } from '../resolve';
+
+const PLUGIN = { name: '@0no-co/graphqlsp', schema: './schema.graphql' };
+
+type Fixture = Record<string, unknown>;
+
+const inFixture = async (files: Fixture, fn: (root: string) => Promise<void>): Promise<void> => {
+  const root = await fs.mkdtemp(path.join(os.tmpdir(), 'gql-tada-resolve-'));
+  try {
+    for (const name of Object.keys(files)) {
+      const filePath = path.join(root, name);
+      await fs.mkdir(path.dirname(filePath), { recursive: true });
+      const contents = files[name];
+      await fs.writeFile(
+        filePath,
+        typeof contents === 'string' ? contents : JSON.stringify(contents, null, 2)
+      );
+    }
+    await fn(root);
+  } finally {
+    await fs.rm(root, { recursive: true, force: true });
+  }
+};
+
+describe('loadConfigs', () => {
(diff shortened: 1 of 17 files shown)


packages/internal/src/__tests__/resolve.test.ts
```diff
@@ -0,0 +1,266 @@
...
+import { describe, it, expect } from 'vitest';
+
+import { loadConfig, loadConfigs } from '../resolve';
+
+const PLUGIN = { name: '@0no-co/graphqlsp', schema: './schema.graphql' };
+
+type Fixture = Record<string, unknown>;
+
+const inFixture = async (files: Fixture, fn: (root: string) => Promise<void>): Promise<void> => {
+  const root = await fs.mkdtemp(path.join(os.tmpdir(), 'gql-tada-resolve-'));
```
kitten: fails to resolve the realpath, but a later `require.resolve` will resolve it, so would be best to pre-resolve

packages/internal/src/resolve.ts
```diff
@@ -69,13 +69,67 @@ const getPluginConfig = (tsconfig: TsConfigJson | null): Record<string, unknown>
...
+
+  return null;
+};
+
 export interface LoadConfigResult {
   pluginConfig: Record<string, unknown>;
+  /** The config file in which the plugin entry was found (may be an `extends` base file). */
   configPath: string;
+  /** The project's own tsconfig.json file, before `extends` resolution. */
+  tsconfigPath: string;
```
kitten, line 134: Ah, funnily enough, also addressed in #545. Would probably need one to rebase or incorporate the other, then we can merge both and hand-test

kitten (approved):

PR #542, opened 2026-06-12
[codex] Improve parser type dispatch performance

JoviDeCroock:
## Summary

This tightens the type-level parser dispatch paths in `src/parser.ts` so the checker avoids several parser branches when the next token already rules them out.

Changes include:
- dispatching `takeValue` from a single head-token match instead of repeatedly matching the full input tuple
- skipping directive parsing unless the next token is a directive
- skipping selection-field vs fragment-spread branches based on the next token
- avoiding document and variable-definition parser branches when the leading token cannot match them

## Impact

The public AST shape is unchanged. The improvement is aimed at reducing TypeScript checker work for larger GraphQL documents while keeping small documents roughly neutral.

## Benchmarks

Baseline captured before editing on this machine:
- `parse simple document`: `15.4045ms` mean
- `parse kitchensink document`: `130.45ms` mean

Final repeated runs with the same parser bench:
- `parse simple document`: `15.7576ms`, `15.7826ms`, `15.6316ms` mean
- `parse kitchensink document`: `107.91ms`, `113.22ms`, `108.39ms` mean

Command:
```sh
./node_modules/.bin/vitest bench src/__tests__/parser.bench.ts --run --typecheck.enabled=false
```

### DIFF
--- a/.changeset/warm-parsers-dance.md
+++ b/.changeset/warm-parsers-dance.md
@@ -1,0 +1,5 @@
+---
+'gql.tada': patch
+---
+
+Improve the type-level GraphQL parser's TypeScript checker performance by dispatching from leading tokens and skipping parser branches that cannot match.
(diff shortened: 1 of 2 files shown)


kitten (approved):
ah didn't think of carrying this the whole way through

PR #573, opened 2026-07-09
fix(internal): Parse tsconfig.json without the TypeScript compiler API

### LINKED ISSUE #572: CLI crashes with `ts.parseConfigFileTextToJson is not a function` on TypeScript 7
### Describe the bug

In a project with TypeScript 7 installed, `gql.tada generate-output` crashes before doing any work:

```
TypeError: ts.parseConfigFileTextToJson is not a function
```

TypeScript 7 no longer exposes the legacy JS compiler API from the `typescript` package's top-level export (a stable programmatic API is planned for 7.1+), and `@gql.tada/internal`'s `readTSConfigFile` uses `ts.parseConfigFileTextToJson` to read `tsconfig.json`. So loading the plugin config fails even though this path only needs to parse JSONC, not run a compiler. `gql.tada generate schema` hits the same crash when it infers the output location from config.

Commands that build a TypeScript program (`check`, `scan`, `turbo`, `generate-persisted`, `doctor`) also depend on the removed API, but that's a larger migration — this issue is only about the config-loading crash. I have a fix for it and will open a PR shortly.

### Reproduction

```sh
mkdir ts7-repro && cd ts7-repro
npm init -y
npm install typescript@7.0.2 gql.tada graphql
cat > tsconfig.json << 'EOF'
{
  "compilerOptions": {
    "strict": true,
    "plugins": [
      {
        "name": "gql.tada/ts-plugin",
        "schema": "./schema.graphql",
        "tadaOutputLocation": "./src/graphql-env.d.ts"
      }
    ]
  }
}
EOF
echo 'type Query { hello: String }' > schema.graphql
npx gql.tada generate-output
```

Expected: `src/graphql-env.d.ts` is written.
Actual: `TypeError: ts.parseConfigFileTextToJson is not a function`

The same steps work with `typescript@5` or `typescript@6.0.0-beta`.

### gql.tada version

`gql.tada@1.11.2` (`typescript@7.0.2`)

### Validations

byrnehollander:
Resolves #572

## Summary

TypeScript 7 no longer exposes the legacy JS compiler API from the `typescript` package's top-level export, so loading the GraphQLSP plugin config crashes with `TypeError: ts.parseConfigFileTextToJson is not a function`. This breaks `gql.tada generate-output` (and `generate schema` when it infers the output location) before any schema work happens.

Reading `tsconfig.json` only needs a JSONC parser, not a compiler, so this swaps `ts.parseConfigFileTextToJson` for `jsonc-parser`.

This intentionally isn't full TS7 support: commands that create a TypeScript program (`check`, `scan`, `turbo`, `generate-persisted`, `doctor`) still use the legacy compiler API. TypeScript 7.0 [ships no programmatic API](https://devblogs.microsoft.com/typescript/announcing-typescript-7-0/) (a new, different one is expected in 7.1), and Microsoft publishes `@typescript/typescript6` as the interim path for tools that need the old API — those commands could load it as a compatibility bridge until the new API is worth porting to. Happy to write that up in a follow-up issue if there's interest.

## Set of changes

- `@gql.tada/internal`: `readTSConfigFile` now parses tsconfig contents with `jsonc-parser` (Microsoft-maintained, matches tsconfig's JSONC dialect). Comments and trailing commas still work; a UTF-8 BOM is stripped and blank files are treated as empty configs, matching the old TypeScript parser's behavior. Parse failures throw a `TadaError` with file path and line/column.
- New `jsonc-parser` dependency in `@gql.tada/internal`, kept external by the rollup build.
- Regression tests for JSONC configs, BOM-prefixed files, and empty referenced configs.
- Patch changeset for `@gql.tada/internal`, `@gql.tada/cli-utils`, and `gql.tada`.

Not a breaking change.

### DIFF
--- a/packages/internal/src/resolve.ts
+++ b/packages/internal/src/resolve.ts
@@ -1,12 +1,12 @@
-import ts from 'typescript';
 import * as path from 'node:path';
 import * as fs from 'node:fs/promises';
 import { createRequire } from 'node:module';
 import type { Stats } from 'node:fs';
 import type { TsConfigJson } from 'type-fest';
+import { parse, printParseErrorCode, type ParseError } from 'jsonc-parser';
 
 import { cwd, maybeRelative } from './helpers';
-import { TSError, TadaError } from './errors';
+import { TadaError } from './errors';
 
 const TSCONFIG = 'tsconfig.json';
 
@@ -43,10 +43,26 @@ const toTSConfigPath = (tsconfigPath: string): string =>
 
 export const readTSConfigFile = async (filePath: string): Promise<TsConfigJson> => {
   const tsconfigPath = toTSConfigPath(filePath);
-  const contents = await fs.readFile(tsconfigPath, 'utf8');
-  const result = ts.parseConfigFileTextToJson(tsconfigPath, contents);
-  if (result.error) throw new TSError(result.error);
-  return result.config || {};
+  let contents = await fs.readFile(tsconfigPath, 'utf8');
+  // Like TypeScript itself, tolerate a UTF-8 BOM
+  if (contents.charCodeAt(0) === 0xfeff) contents = contents.slice(1);
+  const errors: ParseError[] = [];
+  const config = parse(contents, errors, {
+    allowTrailingComma: true,
+    allowEmptyContent: true,
+  }) as TsConfigJson | undefined;
+  if (errors.length) {
+    const error = errors[0];
+    const prefix = contents.slice(0, error.offset);
+    const line = prefix.split(/\r\n?|\n/).length;
+    const col = prefix.length - Math.max(prefix.lastIndexOf('\n'), prefix.lastIndexOf('\r'));
+    throw new TadaError(
+      `Failed to parse ${maybeRelative(tsconfigPath)} at ${line}:${col}: ${printParseErrorCode(
+        error.error
+      )}`
+    );
+  }
+  return config || {};
 };
 
 export const findTSConfigFile = async (targetPath?: string): Promise<string | null> => {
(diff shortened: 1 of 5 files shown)


packages/internal/src/resolve.ts
```diff
@@ -1,12 +1,12 @@
...
 import * as path from 'node:path';
 import * as fs from 'node:fs/promises';
 import { createRequire } from 'node:module';
 import type { Stats } from 'node:fs';
 import type { TsConfigJson } from 'type-fest';
+import { parse, printParseErrorCode, type ParseError } from 'jsonc-parser';
 
 import { cwd, maybeRelative } from './helpers';
-import { TSError, TadaError } from './errors';
+import { TadaError } from './errors';
```
byrnehollander, line 9: `TSError` only exists to wrap a `ts.Diagnostic`, which this path no longer produces — so parse failures now throw `TadaError` like the rest of this module. The CLI logger treats both names the same when rendering errors, and `TSError` itself is still exported from `errors.ts`, so the public API is unchanged.

mighty-peach:
hey @kitten 👋🏻 
is there any chance to get this merged? it is quite a blocked for having ts7 😬 

thanks

JoviDeCroock:
Are you sure that TS7.0 is working after this change, from what I can tell it should not be working given that the Go interface is not set up to be extended nor does it have a way for JS to read the types. See: https://github.com/0no-co/GraphQLSP/issues/421

This PR achieves crash protection but the crash imho is a good way to tell folks that this is in fact not compatible with TS7.0 as they have, knowingly, not added support for things like `gql.tada`, `vue`, ...

